PR #155, opened 2019-04-01
feat(accordion) : add accordion angular 7 component

daanishnasir:
Angular 7 Accordion Component 
## Description

Accordion 
## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


angular/src/lib/accordion/accordion.component.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  @Input() header: string = '';
+  /** @prop Set accordion tab to be expanded | false  */
+  @Input() isExpanded: boolean = false;
+  /** @prop Set the attribute disabled to the accordion tab | false */
+  @Input() disabled: boolean = false;
+  /** @prop Optional accordion tab css class string | '' */
+  @Input() tabClass: string = '';
+  /** @prop Set the height of the AccordionHeader to either the default or 56px | '' */
+  @Input() height: string = '';
```
ArthusLiang: Will it be good to set the type of height to number here?

angular/src/lib/accordion/accordion.component.ts
```diff
@@ -0,0 +1,267 @@
...
+  get headerClasses() {
+    return {
+      ['cui-accordion__header--' + this.height]: this.height,
+      'cui-accordion__header--separator': this.showSeparator,
+      [this.headerClass]: this.headerClass
+    };
+  }
+
+  private isHeightValid = () =>
+    ['', '56'].includes(this.height)
```
ArthusLiang: It's good to use includes for more height options. But since we only have 56 and null options, we may have some more effective way to do the validation.

angular/src/lib/accordion/accordion.component.ts
daanishnasir: agreed, changed to number

angular/src/lib/accordion/accordion.component.ts
daanishnasir: removed this check function and moved it to the get headerClasses()

PR #170, opened 2019-04-12
fix(modal): add example and reorg the files

ArthusLiang:
add modal examples and fix CDK conflict

format

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [x] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


angular/src/lib/modal/examples/modal-default.component.ts
```diff
@@ -0,0 +1,37 @@
...
+        <a (click)='closeModal()'>close</a>
+      </cui-modal-footer>
+    </cui-modal>
+  `
+})
+export class ExampleModalDefaultComponent {
+
+  @ViewChild('modal') modal: ModalComponent;
+
+  constructor() {
```
daanishnasir: I think we should try to find a way to open/close the modal without a ViewChild ref @ArthusLiang. Otherwise its another thing for a user to add to make it work.

ArthusLiang: > Can we see if we can make the modal open/close work without adding viewchild ref

There is a close button in modal-header. I do know if it is good to bind open feature by passing an element to modal. Do you have any other good suggestions?

daanishnasir: Talked about it with Arthus, this ViewChild Ref solution looks ok as the component isn't able to go off a variable change - true/false to open/close the modal.  The other option we talked about was using a service -- discussed how the service should be used internally on the component not an example. @pauljeter

Issue #198, opened 2019-04-29
[ComboBox] add 'ariaLabel' and 'ariaDescribedBy' props for accessibility


AMANDUA:
add 'ariaLabel' and 'ariaDescribedBy' props to ComboBox for accessibility

## Issue Type

Feature
## Component or Pattern Affected

ComboBox
## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Library: React
* Version used: 
* Browser Name and version: 
* Operating System and version: 
* Link to your project:

pauljeter:
Fixed in #199

PR #199, opened 2019-04-29
feat(ComboBox): add ariaLabel and ariaDescribedBy props for accessibility

AMANDUA:
add ariaLabel and ariaDescribedBy props to ComboBox Input feild for accessibility

## Related Issue

https://github.com/momentum-design/momentum-ui/issues/198

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/react/src/lib/Input/index.js
+++ b/react/src/lib/Input/index.js
@@ -121,6 +121,8 @@ class Input extends React.Component {
 
   render() {
     const {
+      ariaLabel,
+      ariaDescribedBy,
       children,
       className,
       clear,
@@ -213,6 +215,8 @@ class Input extends React.Component {
         value={value}
         {...disabled && { disabled }}
         {...htmlId && { id: htmlId }}
+        {...ariaLabel && { 'aria-label': ariaLabel }}
+        {...ariaDescribedBy && { 'aria-describedby': ariaDescribedBy }}
         {...otherProps}
         {...placeholder && { placeholder }}
         {...readOnly && { readOnly }}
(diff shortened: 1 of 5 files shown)


AMANDUA:
Yes @bfbiggs , its possible to pass in input props but in this way

image

i can't pass ```aria-label``` without quotes.

bfbiggs:
If the concern is the quotes and accessibility then I suggest we add those props to the Input itself. Adding to the wrapper components would then require changes to any similar wrapper components eventually. @AMANDUA @pauljeter

AMANDUA:
Agree @bfbiggs , so i will be updating the PR

react/src/lib/Input/index.js
```diff
@@ -260,6 +264,10 @@ class Input extends React.Component {
 }
 
 Input.propTypes = {
+  /** @prop Text to display for blindness accessibility features | '' */
```
bfbiggs: If your default prop is null then it should be the word null on the other side of the pipe.
```
@prop Text to display for blindness accessibility features | null
```

react/src/lib/Input/index.js
```diff
@@ -319,6 +327,8 @@ Input.propTypes = {
 };
 
 Input.defaultProps = {
+  ariaLabel: null,
```
bfbiggs: Also let's keep this alphabetic.

react/src/lib/Input/index.js
```diff
@@ -121,6 +121,8 @@ class Input extends React.Component {
 
   render() {
     const {
+      ariaLabel,
```
bfbiggs: Please alphabetize.

AMANDUA:
Hi @bfbiggs I've updated the code taking care of the comments . Please review

PR #206, opened 2019-05-03
feat(Alert): restructure Alert architecture, remove AlertMeetingContainer

### LINKED ISSUE #110: [Alert] Add timeouts to dismiss alerts
Allow alerts to take in a timeout prop so it can be dismissed after a certain amount of time

## Issue Type
Feature Request

Alert/AlertContainer

1. 

1. 

1. 

1. 

## Your Environment

* Library: 
* Version used: 
* Browser Name and version: 
* Link to your project:

### LINKED ISSUE #122: [Alert] Add onClick handler to alert itself + include accessibility
Currently there's an onHide prop which gets passed to the X button on the Alert. However, all the remaining props are being passed to this X button so there is no way to pass an onClick handler to the outer container of the Alert. 

1) Remaining props should be on the outer container so we can also pass in className and other props
2) Alert should take handle an onClick handler prop (and also handle key downs for accessibility and keyboard navigation through the AlertContainer list)

## Issue Type
Bug Report/Feature Request

Alert/Alert Container

## Expected Behavior
Clicking Alert should trigger an action

## Current Behavior
Clicking Alert does not trigger an action

## Possible Solution
1) Remaining props should be on the outer container so we can also pass in className and other props
2) Alert should take handle an onClick handler prop (and also handle key downs for accessibility and keyboard navigation through the AlertContainer list)

1.
2.
3.
4.

## Your Environment

* Library: 
* Version used: 
* Browser Name and version: 
* Link to your project:

bfbiggs:
…iner

BREAKING CHANGE:

AlertMeetingContainer has been removed
Alert is now passed in as children to AlertContainer
Alert otherProps now passed to outer div
Alert add dismissBtnProps for close button
Alert closable prop now defaults to true
AlertContainer remove orderNewest prop/logic

## Related Issue

Resolves #122 
Resolves #110 
## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


react/src/lib/Alert/examples/Default.js
```diff
@@ -1,36 +1,41 @@
...
           <div className='row'>
             <Button
               ariaLabel='Click to Open'
-              onClick={() => alertContainer.info(
-                'Alert',
-                this.state.alertMessage,
-                () => alert('onHide info'),
-                { ariaLabel: 'Close Alert' }
-              )}
+              onClick={() => this.setState({ show: true })}
```
aimeex3: would a more realistic example be adding the alert to an array state onClick, and then pop it off the array on dismiss?

react/src/lib/Alert/examples/Default.js
```diff
@@ -1,36 +1,41 @@
 import React from 'react';
 import {
+  Alert,
   AlertContainer,
   Button
 } from '@momentum-ui/react';
 export default class AlertDefault extends React.PureComponent {
   state = {
-    alertMessage: 'Who\'s awesome?  You are!'
+    alertMessage: 'Who\'s awesome?  You are!',
```
aimeex3: not using state.alertMessage in the examples anymore

react/src/lib/Alert/examples/Default.js
bfbiggs: You could still keep track of an array in your app, and just leave all alerts with the show prop as true by default.  This is specifically for the example.

react/src/lib/Alert/examples/Default.js
bfbiggs: This was an artifact from when all the examples were in one file and the original author didn't want to write the same message.

PR #240, opened 2019-05-21
feat (cards) angular & core

mathsusan:
Adding cards to angular and css to core 

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

- [x ] I have read the **CONTRIBUTING** document.
- [ x] All new and existing tests passed.


angular/src/lib/card/examples/card-content.component.ts
```diff
@@ -0,0 +1,62 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'example-card-content',
+  template: `
+    <md-card [ngStyle]="{'width': '300px'}">
```
daanishnasir: Will a width prop work instead of ngStyle

mathsusan: fixed

bfbiggs:
The reason discussed just using ngStyle is so that other style props could be changed as well, similar to React where we allow ...props in the majority of our components instead of limiting it to just one property. @daanishnasir @mathsusan

core/scss/components/card/settings.scss
```diff
@@ -1,6 +1,9 @@
+// THESE DON'T WORK
```
bfbiggs: What's the meaning of this comment?

core/scss/components/card/card.scss
```diff
@@ -1,1073 +1,27 @@
...
-        transition: transform 0.6s ease;
-      }
-
-      .card-menu-heading {
-        @include vr-spacing(p, 1.5);
-
+    position: relative;
+    display: flex;
+    color: $text-color;
+    vertical-align: top;
```
bfbiggs: Should this be using flex prop instead of vertical-align? (justify/align)

core/scss/components/card/card.scss
```diff
@@ -1,1073 +1,27 @@
...
+    background-color: $card-color;
+    border-radius: $card-border-radius;
+    box-shadow: 0px 1px 2px 1px rgba(52, 53, 55, 0.18);
+    flex-direction: column;
+      .md-card-section {
```
bfbiggs: What's the reason for card-section vs card-full-section? Seems like it could be consolidated with a prop/input.

bfbiggs:
Also need to add test folder/test for scss.

core/scss/components/card/settings.scss
mathsusan: we need to talk about this .. if I make changes here nothing happens.  Glad you caught the comment because I forgot about it

core/scss/components/card/card.scss
mathsusan: that should not be there.

angular/src/lib/card/examples/card-content.component.ts
```diff
@@ -0,0 +1,63 @@
...
+      display: flex;
+      justify-content: flex-end;
+      align-items: baseline;
+    }
+    .footer-link {
+      margin-right: 16px;
+    }
+    `
+  ]
+})
```
daanishnasir: Can we add the styling of cards to the core .scss file instead of doing it here? I would think react would need this when its created; unless it's specific to angular. @pauljeter

mathsusan: card only do the padding css.. all the things inside would need to do this themselves.  We can't guess all the different things they might want to do inside a card.

core/scss/components/card/examples/card.html
```diff
@@ -0,0 +1,19 @@
+
```
pauljeter: can you rename this to default.html? Also, can we add the same examples that you did in Angular (content and content2)?

PR #255, opened 2019-06-04
fix(popover): add popoverTrigger Input

mathsusan:
## Description

adding popoverTrigger which was not there

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

- [ x] I have read the **CONTRIBUTING** document.
- [ x] I have added tests to cover my changes.
- [x ] All new and existing tests passed.

### DIFF
--- a/angular/src/lib/popover/tests/popover.component.spec.ts
+++ b/angular/src/lib/popover/tests/popover.component.spec.ts
@@ -37,7 +37,7 @@ describe('Popover Test', () => {
     fixture = TestBed.createComponent(TestPopoverComponent);
     component = fixture.componentInstance;
     testEl = fixture.debugElement.query(By.css('md-badge'));
-    testEl.triggerEventHandler('mouseenter', null);
+    testEl.triggerEventHandler('click', null);
     fixture.detectChanges();
   }));
 
@@ -45,7 +45,7 @@ describe('Popover Test', () => {
     expect(overlayContainerEl).toMatchSnapshot();
   });
 
-  it('popover should appear when hovering over badge NOTE: it uses tooltip', () => {
+  it('popover should appear when click over badge', () => {
 
     const popoverEl = overlayContainerEl.querySelector('md-tooltip');
     expect(popoverEl).not.toBeNull();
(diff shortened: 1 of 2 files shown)


angular/src/lib/popover/tests/popover.component.spec.ts
```diff
@@ -37,15 +37,15 @@ describe('Popover Test', () => {
...
+    testEl.triggerEventHandler('click', null);
     fixture.detectChanges();
   }));
 
   it('should match snapshot', () => {
     expect(overlayContainerEl).toMatchSnapshot();
   });
 
-  it('popover should appear when hovering over badge NOTE: it uses tooltip', () => {
+  it('popover should appear when cklick over badge', () => {
```
bfbiggs: spelling

angular/src/docs/app/playground/playground.component.ts
```diff
@@ -6,6 +6,7 @@ import { Component, OnInit } from '@angular/core';
     <div className="row">
       <div className="example-spacing">
         <!-- Insert Example Here -->
+        <example-popover-trigger></example-popover-trigger>
```
bfbiggs: clean up playground

bfbiggs:
Can you rename your commit message to 'fix(popover): add popoverTrigger Input'

mathsusan:
> I tested it in the playground, we may need an additional prop or something that will keep the popover open when clicking inside of the popover. For instance, we may have a list/selects inside the popover that would need to stay visible after clicking. Clicking outside of the popover will close the popover.
> 
> https://drive.google.com/file/d/1bXlQytGUzwf-UsZwIHwHv_8PwhJU2JSf/view
> 
> Pretty much how the click example works in React.
> 
> https://stage.momentum.design/components/popover/code
> @mathsusan @pauljeter @bfbiggs

This definitely sounds like a feature request it was not part of the original requirements and should be put on the backlog.

bfbiggs:
I think the intent of the task was to allow it to have the same functionality as React. The props Daani is referencing are probably closeOnClick and allowClickAway that actually live in the EventOverlay component in React but since we're using the CDK in Angular, probably need to live in the Popover. Let's merge and we can revisit any additional needs in planning.

PR #256, opened 2019-06-04
Sidebar toggle

bfbiggs:
## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


react/src/lib/Sidebar/index.js
```diff
@@ -48,14 +49,17 @@ class Sidebar extends React.Component {
 
   render() {
     const {
+      buttonProps,
       children,
       className,
       expandable,
```
pauljeter: I think that we should use `collapsable` as the prop name since we are defaulting to expanded.

react/src/lib/Sidebar/index.js
```diff
@@ -48,14 +49,17 @@ class Sidebar extends React.Component {
 
   render() {
     const {
+      buttonProps,
       children,
       className,
-      expandable,
+      collapsable,
```
mathsusan: does this mean you can turn off the collapsing?  or should the variable be collapsed??  mean it's collapsed or not?

pauljeter: This prop is to turn on/off collapsing. i.e. on momentum.design, the sidebar is not collapsable but in Atlas it is.

PR #259, opened 2019-06-06
feat(Lightbox): add Angular Lightbox component

wzpo:
lightbox

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


angular/src/lib/lightbox/examples/multiple.component.ts
```diff
@@ -0,0 +1,45 @@
...
+        thumb: 'https://momentum.design/cisco-webex-lockup-blue.png',
+      }],
+      onDownload: () => {
+        this.lightbox.downloading = true;
+        setTimeout(() => this.lightbox.downloading = false, 2000);
+      },
+      onClose: () => {},
+    });
+  }
+}
```
mathsusan: is it possible to use the light box with html or I have to write this script to do it only?  Seems like it would be a lot more user friendly to be able to use it with html

wzpo: Currently, we cannot use html with props to do it. I think it's more flexible and the code is easy to read to use service with config script.

PR #263, opened 2019-06-10
Event overlay scroll parent

### LINKED ISSUE #204: [Popover] Handle isDynamic and isContained (including custom padding prop)
When adding isContained, depending on the position of the avatar, the popover stretches. I tried using isDyanmic too and that didn't work. We want the popover to show within the browser window so if it's near the top or near the bottom, you can still see the popover contents. It doesn't need to be contained within the space list (which is what isContained would do I believe) I think isDynamic might be the correct prop to use, but doesn't look like it works correctly either

For the dynamic popover, is it possible to add a custom margin like 12px margin from the edge so that it doesn't go to the edge of the screen?

## Issue Type
Bug Report

Popover

## Expected Behavior
If popover is triggered near the button of the screen and isDynamic is set to true, it should dynamically open upwards and show the full contents of the popover. Same logic applies if it's near the top of the app.

## Current Behavior

if using isContained

1. 
1. 
1. 
1. 

## Your Environment

* Library: 
* Version used: 
* Browser Name and version: 
* Link to your project:

bfbiggs:
## Related Issue

Resolves #204 
## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


react/src/lib/Select/index.js
```diff
@@ -90,6 +90,7 @@ class Select extends React.Component {
 
   render() {
     const {
+      btnClassName,
```
pauljeter: Should this be `buttonClassName`? Other than the old AngularJS code, we don't abbreviate button in our components.

bfbiggs: We have @pauljeter:
- Alerts:   dismissBtnProps: PropTypes.object,
- ModalHeader: closeBtnProps: PropTypes.object,

After our discussion, buttonProps wouldn't necessarily work since we are already applying a className to the button.  Can discuss in detail tomorrow.

PR #268, opened 2019-06-18
fix(tooltip): add condition to include tooltip text

mathsusan:
## Description

extra padding was in the popover when I was using the template.  Also, discovered an exception sometimes when using the tooltip or popover

## Motivation and Context

extra padding was in the popover when I was using the template.  Also, discovered an exception sometimes when using the tooltip or popover

## Screenshots:
**Before** (If applicable):

before

**After**:

after

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

- [x ] I have read the **CONTRIBUTING** document.
- [ x] All new and existing tests passed.


angular/src/lib/tooltip/tooltip.directive.ts
```diff
@@ -153,12 +153,14 @@ export class TooltipDirective implements OnInit {
...
 
   public close() {
-      setTimeout(() => {
-        this.overlayRef.detach();
-        if (this.tooltipRef ) {
-          this.tooltipRef.destroy();
-          this.tooltipRef = null;
-        }
-      }, this.delay);
+      // setTimeout(() => {
```
bfbiggs: Revert possibly?

bfbiggs:
Not sure if you saw my comment on the other pr commit message: 
1) need the colon after the parens for sem-ver to catch patch fix
2) based on the file changed (tooltip) it seems like it would be something like fix(tooltip): add condition to include tooltip text

https://github.com/momentum-design/momentum-ui/blob/master/CONTRIBUTING.md

PR #275, opened 2019-06-20
fix(card): core, angularjs and angular fixes

mathsusan:
BREAKING CHANGE:

angularng cards will be broken

## Description

fix the core css to make sure cards css for angularng don't conflict

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [x ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

- [ x] I have read the **CONTRIBUTING** document.
- [x ] All new and existing tests passed.

### DIFF
--- a/core/scss/ng/card.scss
+++ b/core/scss/ng/card.scss
@@ -1,5 +1,5 @@
 @include exports('ng-card') {
-    .md-card-container {
+    .md-ng1-card-container {
       display: flex;
       width: 100%;
       padding-right: $page-side-padding - $card-padding-horizontal;
@@ -18,8 +18,8 @@
       }
     }
 
-    .md-card-layout {
-      &.md-card-block {
+    .md-ng1-card-layout {
+      &.md-ng1-card-block {
         @include vr-spacing(ph, 1);
         @include grid-row($behavior: nest);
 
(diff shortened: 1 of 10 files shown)


core/scss/ng/card.scss
```diff
@@ -18,23 +18,23 @@
       }
     }
 
-    .md-card-layout {
-      &.md-card-block {
+    .md-ng1-card-layout {
+      &.md-ng1-card-block {
```
bfbiggs: I think we only need to add ng1 to the top levels. This will reduce adoption time for html users as well.

daanishnasir: Good idea

.md-ng1-card {
   &--block

mathsusan:
I don't think we want to change the other cards until this is in?  Plus you said to separate the two commits

core/scss/ng/card.scss
mathsusan: made the change on the top .. not touching the rest as I have no way to test this change

bfbiggs:
Couple Issues:
- initial commit should be fix(card) because in scss/ng the file is singular
- core examples need to be updated to remove the v2
- angularjs cards component needs to be updated to the new class

angularjs/src/lib/components/cardnumber/cardnumber.component.ts
```diff
@@ -14,7 +14,7 @@ export class CardNumberCtrl implements ng.IComponentController {
 export class CardNumberComponent implements ng.IComponentOptions {
   public controller = CardNumberCtrl;
   public template = `
-  <div class="md-card-number">
+  <div class="md-ng1-card-number">
```
bfbiggs: No need to change this since it wasn't modified in the scss/ng/card-number.scss.

PR #291, opened 2019-07-10
fix(button): change throw error to console.warn for missing aria-label

daanishnasir:
## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


angular/src/lib/button/button.component.ts
```diff
@@ -155,10 +155,9 @@ export class ButtonComponent implements AfterViewInit {
 
   private checkAriaLabel() {
     const ariaLabel = this.getAttr('aria-label');
-    if (!ariaLabel && this.el.nativeElement.children.length > 0) {
-      throw new Error(
-        'md-button: content is not a string, you must add an "ariaLabel" for accessibility.'
-      );
+
+    if (!ariaLabel && this.el.nativeElement) {
+      console.warn('md-button: content is not a string, you must add an "aria-label" for accessibility.');
```
brspence: `this.el.nativeElement` is this component's element. Won't that always be true and warn if there is no `aria-label` attribute?

Aria label is used to label something when there is no visible text. Is the purpose here to programmatically identify when aria label needs to be applied or always force/encourage the implementor to provide an aria label even if they don't need to (in which case `nativeElement` check is unnecessary)? Not sure how you'd guarantee that some child node doesn't have inner text that would be visible.

bfbiggs: Agreed - not sure why the additional logic was added.

PR #295, opened 2019-07-22
fix(List): fix active/focus context on list

aimeex3:
## Description
active/focus are only set on mount. If data is changing in the list, active/focus items do not get updated so keyboard navigation does not work correctly. 

## Motivation and Context

active/focus are only set on mount. If data is changing in the list, active/focus items do not get updated so keyboard navigation does not work correctly. 

## How Has This Been Tested?

built the package and copied it over to web-client repo. Tested keyboard navigation in the app

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


react/src/lib/ListItem/index.js
```diff
@@ -200,8 +200,8 @@ class ListItem extends React.Component {
               contextProps.id = this.props.id || id;
               contextProps.uniqueKey = navKey || contextProps.id;
               contextProps.type = type || (listContext && listContext.type);
-              contextProps.focus = focus || (listContext && listContext.focus === contextProps.uniqueKey);
-              contextProps.active = active || (listContext && listContext.active === contextProps.uniqueKey);
+              contextProps.focus = focus || (listContext && makeEventKey(listContext.focus) === contextProps.uniqueKey);
```
bfbiggs: This code is problematic because makeEventKey forces kebab case when the id may not be kebab.  It's meant to be ran only if the eventKey is provided.  I'm sure there's a better way to handle this, but currently this breaks the code when relying on the id from react-uid.

react/src/lib/List/index.js
```diff
@@ -52,6 +52,9 @@ class List extends React.Component {
 
   componentDidUpdate(prevProps, prevState) {
     const { listContext } = this.state;
+    if (this.props.shouldFocusActive && (prevProps.focus !== this.props.focus || prevProps.active !== this.props.active)) {
```
bfbiggs: Can we destructure the this.props props to shorten this up a bit? Other than that looks great.

PR #307, opened 2019-08-01
feat(angularjs)-ng-transclude for page header right

yeonskim:
Add transclude for placing the content in page header
## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/angularjs/src/lib/components/pageheader/pageheader.component.ts
+++ b/angularjs/src/lib/components/pageheader/pageheader.component.ts
@@ -37,14 +37,13 @@ export class PageHeader implements ng.IComponentOptions {
         <i ng-if="$ctrl.back" class="icon icon-arrow-left_16" role="link" ng-click="$ctrl.goTo()" aria-label="{{::$ctrl.backAriaLabel}}"></i>
         <h2 class="page-header__title">{{$ctrl.titleName}}</h2>
       </div>
-      <div class="page-header__right">
+      <div class="page-header__right" ng-transclude="pageheaderright">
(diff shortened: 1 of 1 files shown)


angularjs/src/lib/components/pageheader/pageheader.component.ts
```diff
@@ -44,10 +44,14 @@ export class PageHeader implements ng.IComponentOptions {
               <a href class="h4" ui-sref="{{::tab.state}}" ng-class="{{::tab.class}}" id="{{::tab.id}}" role="tab">{{tab.title}}</a>
             </li>
           </ul>
+          <span class="page-header__right__content" ng-transclude="pageheaderright">
```
brspence: Can the transclusion be on the actual `<div class="page-header__right">` element?
```html
<div class="page-header__right" ng-transclude="pageheaderright">
  <div class="md-tab md-tab--tabs md-tab--nav">
  <ul> ... </ul>
  </div>
</div>
```

That way when no transclusion is provided, the tab structure is still default. But providing an `<md-page-header-right>` transclusion would replace the entire `page-header__right` content.

Think our UX only had the toggle switch on the overview page which has no tab content. If we needed to put the toggle switch on other page headers with tabs, would need a transclusion point *within* the `page-header__right`, but preceding the tabs. Can probably just go with the full `page-header__right` replacement since that is our primary use case.

PR #325, opened 2019-08-16
feat(input): add warning built-in angular validator, update example

daanishnasir:
- Added Warning Custom Validator 
- Warning and Error can now be used together when using Angular Form Validators
- cascaded error css down below warning css so error styling will override warning styling if both classnames are ever together

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


daanishnasir:
@sani12

daanishnasir:
```  
errorObj = {
    required: 'This field is required',
    minlength: 'This field should be more than 8 chars',
    maxlength: 'This field cant be more than 25 chars',
    email: 'Not a valid email',
    custom: 'Custom Validator Error - input does not start with "custom"', // create custom validator like below
    warning: 'Warning Validator - input does not contain a number' // create warning validator like below
  };

File: core/scss/components/input/input.scss [hunk 1]
```diff
@@ -131,39 +131,39 @@
       }
     }
 
-    &.error {
+    &.warning {
```
bfbiggs: Was this just a compilation change so that error was below warning? Not a big deal as we transition to the new input but we'd probably want to filter the possible classes to apply in the javascript instead of applying both classes if possible.

daanishnasir: yeah just to keep it below warning

pauljeter: I agree that we should try to only apply one className with the JS if possible, however the order change should also be applied for the priority *IF* both classes exist on the same input.

PR #347, opened 2019-08-31
Avatar Angular Element

akoushke:
### For Demo Only

 I was able to create the angular element wrap around the angular element.


pauljeter:
We're not going to use Angular Elements for web components. Still exploring our options.

PR #357, opened 2019-09-10
fix(slider): Make slider accessible

### LINKED ISSUE #94: Accessibility for Slider
## Issue Type
Bug Report 

Slider

## Expected Behavior
I should be able to hear what value is when I interact with the slider, similar to https://material.angular.io/components/slider/examples

## Current Behavior
Have no clue what value does the slider has

## Possible Solution
Similar to angular material slider

1. 
1. 
1. 
1. 

## Your Environment

* Library: 
* Version used: 
* Browser Name and version: 
* Link to your project:

jbirori:
## Description
Included aria attributes to the div wrapping the slider to make it more accessible. If the slider has one pointer then it uses the `aria-valuenow` attribute to share the current value. If the slider has two pointers it uses the `aria-valuetext` attribute to share the value of both pointers in the following format:
 `"Low is Math.min(sliderLow, sliderHigh), high is Math.max(sliderLow, sliderHigh)"`
The min/max operations handle the case where the two pointers can cross each other.

**Alternate Solution**: In testing with a screen reader, I found that it actually makes more sense to add the aria attributes directly to the pointers so that the user can be updated on movements of the pointers. However, this would require adding in one or more props to the SliderPointer component to make it aware of its current value. Any thoughts on which approach should be prioritized?

## Related Issue
Resolves #94 (for React)

## Motivation and Context
The slider currently has no accessibility features.

## How Has This Been Tested?
I updated the jest tests, used Chrome Developer tools, and also used a screen reader to test the changes.

## Screenshots:
**Before** (If applicable):
There were no aria attributes

**After**:
`aria-valuenow` for one slider pointer:
Screen Shot 2019-09-10 at 3 42 38 PM

`aria-valuetext` for two slider pointers:
Screen Shot 2019-09-10 at 3 41 29 PM

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

- [x] I have [updated] tests to cover my changes.


react/src/lib/Slider/index.js
```diff
@@ -155,10 +155,24 @@ class Slider extends React.Component {
     };
 
     return (
-      <div className={
-        `md-slider ${className}` +
-        `${(disabled && ` md-slider--disabled`) || ''}` +
-        `${(className && ` ${className}`) || ''}`
```
jbirori: Also this is a bit redundant because className is already added 2 lines up. I suggest we remove that and stay with this conditional line.

react/src/lib/Slider/index.js
```diff
@@ -155,10 +155,24 @@ class Slider extends React.Component {
...
+      <div 
+        className={
+          `md-slider ${className}` +
```
bfbiggs: Agree with removing class name here as well because it also adds unnecessary spaces.

PR #362, opened 2019-09-12
Date picker accessibility react

### LINKED ISSUE #94: Accessibility for Slider

jbirori:
## Description
Added aria attributes to the navigation arrows and to the individual days of the popup calendar. For the navigation arrows I labeled them with the appropriate months that they would be associated with (e.g. "next month, October"). As for the individual days, they were previously labelled with just the number that they represented and I swapped that out for a more explicit date (e.g. "Thursday, September 12th 2019"). 

## Related Issue
Resolves #94 

## Motivation and Context
The accessibility features for the current calendar date picker are not very helpful or non-existent.

## How Has This Been Tested?
I updated the jest tests, used Chrome Developer tools, and also used a screen reader to test the changes.

## Screenshots:
**Before** (If applicable):

**After**:
**Individual Day Aria Label**
Screen Shot 2019-09-12 at 3 11 27 PM

 **Previous Month Aria Label**
Screen Shot 2019-09-12 at 3 11 55 PM

Next Month Aria Label
Screen Shot 2019-09-12 at 3 12 10 PM

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

- [x] I have [updated] tests to cover my changes.


react/src/lib/DatePicker/DatePickerCalendar/index.js
```diff
@@ -87,7 +87,7 @@ class DatePickerCalendar extends React.Component {
       const allPrevDaysDisabled = shouldPrevMonthDisable(date, minDate);
       return (
         <Icon
-          ariaLabel={previousArialLabel}
+          ariaLabel={`${previousArialLabel} month, ${subtractMonths(date.clone(), 1).format("MMMM")}`}
```
bfbiggs: My concern here is internationalization.  If we don't give them control to replace the string completely than month will remain month always.  Perhaps we could have an internal variable that sets this value if previousAriaLabel prop isn't present?

react/src/lib/DatePicker/DatePickerDay/index.js
```diff
@@ -45,11 +45,13 @@ class DatePickerDay extends React.Component {
           `${(isOutsideMonth && ` md-datepicker__day--outside-month`) || ''}`
         }
         onClick={this.handleClick}
-        ariaLabel={`${getDate(day)}`}
+        ariaLabel={`${day.format("dddd, MMMM Do YYYY")}`}
         aria-selected={isSelected}
         tabIndex={-1}
       >
-        {getDate(day)}
+        <div aria-label={day.format("dddd, MMMM Do YYYY")}>
```
bfbiggs: Believe this still needs to be ariaLabel?

jbirori: I initially thought so as well but a flag was thrown up when compiling. Found this in the [React documentation](https://reactjs.org/docs/accessibility.html):

> Note that all aria-* HTML attributes are fully supported in JSX. Whereas most DOM properties and attributes in React are camelCased, these attributes should be hyphen-cased (also known as kebab-case, lisp-case, etc) as they are in plain HTML

PR #413, opened 2019-10-08
feat(core): background updates for CH

mathsusan:
changing backgrounds for control hub refresh
## Description

adding a class for control hub refresh color changes
## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/core/scss/ng/page-header-admin.scss
+++ b/core/scss/ng/page-header-admin.scss
@@ -35,5 +35,11 @@
         margin-right: $page-header-icon-margin-right;
       }
     }
+    .control-hub-refresh {
+      .page-header{
+        background-color: $md-gray-10;
+        border-bottom: none;
+      }
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


core/scss/ng/page-header-admin.scss
```diff
@@ -35,5 +35,11 @@
         margin-right: $page-header-icon-margin-right;
       }
     }
+    .control-hub-refresh {
```
Teamop: [question] don't know any context here, but seems this is only specific to CH, why not just add the style in CH? Cause from my understanding, I think `momentum-ui` should be more general instead of correlated with some products. thanks

mathsusan: Not disagreeing at all but it turns out the entire file is specific to CH and needs moved, that is on the backlog somewhere

Teamop: I see, thanks @mathsusan

PR #414, opened 2019-10-08
fix(topbar-mobile): input not able to use 'SPACE' key on topbar

daanishnasir:
Using an input on topbar would close when trying to use 'space' key. Removed space listener. 

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


angular/src/lib/topbar/topbar-mobile.component.ts
```diff
@@ -64,7 +64,7 @@ export class TopbarMobileComponent {
   @HostListener('document:keydown', ['$event']) handleKeydown(
     e: KeyboardEvent
   ) {
-    if (e.code === 'Space' || e.code === 'Enter') {
+    if (e.code === 'Enter') {
```
bfbiggs: Per our discussion a while back...we know .code won't work in ie11 unless our polyfill is taking care of it.

https://caniuse.com/#search=.code

d-silvas:
Hey, sorry for commenting on this old issue. There are cases where we still need to use the Enter key. For example: a user is typing on an input text box. We'd like to submit the form when the user presses Enter.

As the current topbar is listening for `'document:keydown'` events, it blocks Enter key press events on the whole document. Is there a solution to fix this?

PR #416, opened 2019-10-08
feat(angular):  tooltip add more functionality

mathsusan:
add a hooks to allow the tooltip to stay open until click outside and or allow the user to move their mouse into the tooltip before closing
## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


angular/src/lib/tooltip/tooltip-container.component.ts
```diff
@@ -37,18 +46,35 @@ export class TooltipContainerComponent implements OnInit {
...
   /** @prop Sets the id for screen reader */
   @Input() id: string;
 
-   /** @prop Sets the max-width */
-   @Input() maxWidth: number;
+  /** @prop Sets the max-width */
+  @Input() maxWidth: number;
+
+  @Output() mouseleaveEvent = new EventEmitter();
+  @Output() mouseenterEvent = new EventEmitter();
```
daanishnasir: should we use camelcase wording mouseLeaveEvent, mouseEnterEvent?

pauljeter: I'm ok with not camelCasing these since their browser api counterparts are not camelCased. @bfbiggs?

angular/src/lib/tooltip/tooltip.directive.ts
```diff
@@ -42,16 +42,22 @@ export class TooltipDirective implements OnDestroy {
  /** @prop Sets the offset of the tooltip from the host */
   @Input() offset: number = 5;
 
-   /** @prop Sets the maxwidth of the tooltip */
-   @Input() maxWidth: number = 200;
+  /** @prop Sets the maxwidth of the tooltip */
+  @Input() maxWidth: number = 200;
+
+  /** @prop Allows tooltip to stay open until you click outside | false */
+  @Input() closeOnClick: boolean = false;
```
pauljeter: Should this only be a prop on the Popover? We need to define the UX differences or similarities between Popover and Tooltip

daanishnasir: I would think both should have the capability

mathsusan: if so we need a dark popover

PR #425, opened 2019-10-14
fix(DatePicker): update to new context api

jbirori:
## Description

Update the DatePicker component to the new context api.

## Motivation and Context
DatePicker currently using deprecated api.

## How Has This Been Tested?
Tests have been updated and are all passing.

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have [updated] tests to cover my changes.
- [x] All new and existing tests passed.

PR #434, opened 2019-10-19
fix(DatePicker): make date picker more accessible

### LINKED ISSUE #428: Daterpicker Accessibility
Currently, Datepicker component does not work with screen reader, since the focus is never actually in the component, and instead CSS is used to highlight elements in the calendar.

## Issue Type
Bug Report

Datepicker

## Expected Behavior
Datepicker should be actually accessible via the keyboard i.e. the components in Datepicker should receive focus and should be the active element, only then will the screen reader read for that element. Arrow keys should be able to navigate the calendar, and tab/shift tab should navigate to and from the next/previous month buttons.
[MeetingsPanelP0_7.pdf](https://github.com/momentum-design/momentum-ui/files/3735313/MeetingsPanelP0_7.pdf)

## Current Behavior
Currently, CSS is used to highlight dates to mimic focus, but since they are not the active elements, the screen reader does not read for those elements. Pressing arrow keys after opening the calendar end up closing the calendar.

1.
2.
3.
4.

## Your Environment

* Library: React
* Version used: 20.3.0
* Browser Name and version: Chrome 77, Firefox 69
* Operating System and version: Windows 10
* Link to your project:

jbirori:
## Description
1) Added a ref to the DatePickerDay component so that focus can be assigned to the dates when they are pre-selected via the arrow keys

2) Add a prop to Icon that forwards a ref to the underlying Button component. This grants whatever component is using the Icon component access to the underlying button. Things like assigning focus to the button are now possible because of this.

3) Trap the focus within the date picker calendar when it is open so that tab/tab shift navigate between the next/previous month buttons and the arrow keys navigate the days. Focusable elements outside of the calendar can only be focused on when the calendar is closed or the focus is already outside of the calendar.

## Related Issue
Resolves #428 

## Motivation and Context
Screen readers are currently unable to read the date picker calendar because the the days never receive focus and when the calendar is open it does not 'trap' the focus within the calendar.

## How Has This Been Tested?
All tests are passing and Chrome's screen reader was used for testing with the keyboard keys.

## Screenshots:
**Before** (If applicable):
As you can see in the gif below, using the arrow keys doesn't assign focus (orange outline) to the days, it only does so via css. Also, focus is not trapped into the calendar so it's possible to exit the calendar without wanting to.
datepicker_less_accessible

**After**:
Pressing tab or shit tab navigates between the previous and next month buttons only. If tab is pressed twice it does not navigate to the next focusable element outside of the calendar, it instead continues to switch back and forth between the previous and next month buttons. Arrow keys can now navigate the days and assign focus to the days as they are being navigated.
datepicker_more_accessible

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


react/src/lib/DatePicker/DatePickerDay/index.js
```diff
@@ -13,6 +13,18 @@ import moment from 'moment';
...
   static displayName = 'DatePickerDay';
 
+  constructor(props) {
+    super(props);
+    this.dayButton = React.createRef();
+  }
+
+  componentDidUpdate = () => {
+    const { day } = this.props;
+    const { focus } = this.context;
```
jbirori: Will need to update this when #425 is merged

bfbiggs: You could branch off of your PR right?

bfbiggs: Or just include the commits together.

react/src/lib/DatePicker/DatePickerCalendar/index.js
```diff
@@ -179,6 +192,8 @@ DatePickerCalendar.propTypes = {
   minDate: PropTypes.instanceOf(Date),
   /** Sets the format in how the Month is displayed | null */
   monthFormat: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
+  /** Sets which arrow button for the month navigation should have focus */
```
bfbiggs: Need default since this isn't a required prop.

PR #435, opened 2019-10-21
feat(Vue): add Select

wzpo:
image

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


vue/src/app/playground/index.vue
```diff
@@ -1,11 +1,429 @@
 <template>
-  <div class="row">
-    <div class="example-spacing">
-      <!-- Insert Example Here -->
+  <div class="wrap">
```
mathsusan: shouldn't this be left empty?  like we do in the other projects?  and these examples should be in example files

vue/src/lib/alert/examples/default.vue
```diff
@@ -4,7 +4,7 @@
       <div class="row">
         <md-button
           ariaLabel="Click to Open"
-          @click="() => show = true"
+          @click="() => show = !show"
```
mathsusan: this is for select or should some of these changes be in a separate PR?

vue/src/app/playground/index.vue
wzpo: Removed. This is the demo for engineers.

vue/src/lib/alert/examples/default.vue
wzpo: Reversed. Changed for demo last week.

PR #462, opened 2019-11-05
feat(tabs): add underline tab variation

daanishnasir:
BREAKING CHANGE:

Default of tabs is now underline, pass in tabType=‘pills’ for pill variation, module cleanup

## Description

- Removed empty spec files, moved tests and added more tests to a single test file
- Cleaned up modules, only TabsModule is needed now. 
- Add underline tabs as the default variation 
- Add tabType prop for pill variation 
- focus fix and onKeyHandling works

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


angular/src/lib/tabs/tab-header.component.ts
```diff
@@ -1,22 +1,18 @@
...
 
-import { Component, OnInit, Input } from '@angular/core';
+import { Component, Input } from '@angular/core';
 
 @Component({
   selector: 'md-tab-header',
   template: `
     {{ heading }}
-    <div *ngIf="subHeading">{{ subHeading }}</div>
+    <div *ngIf="subHeading" class="md-tab-header--subheading">{{ subHeading }}</div>
```
bfbiggs: Would this make more sense to either be md-tab-header__subheading as it's an element?

angular/src/lib/tabs/tabs.component.ts
```diff
@@ -21,21 +19,22 @@ export type TabType = 'pills';
...
+  /** @prop Type of Tabs | 'pills' */
+  @Input() tabType: string =  '';
+  /** @prop Callback function invoked when user selects a tab | null */
   @Output() whenSelect = new EventEmitter();
 
   @HostBinding('class') get classes(): string {
     return (
       'md-tab' +
       `${(this.tabType && ` md-tab--${this.tabType}`) || ''}` +
+      `${(!this.tabType && ' md-tab--tabs md-tab--nav') || ''}` +
```
bfbiggs: So should this maybe just be the default or underline tabType?  This creates a breaking change for intending to use pills. @pauljeter

pauljeter: It makes more sense for the Tabs default to be this style vs the pill buttons. I am ok with the breaking change

angular/src/lib/tabs/tab-header.component.ts
pauljeter: I do agree that the className here should be `md-tab-header__subheading`

bfbiggs:
Ideally we would knock out react but let's atleast update the HTML examples.

angular/src/lib/tabs/tabs.service.ts
```diff
@@ -84,7 +86,7 @@ export class TabsService {
   }
 
   public setFocusLast = () => {
-    this.setFocus(this.Tabs - 1);
+    this.setFocus(this.Tabs);
```
ArthusLiang: Why change this default value to -1 ? Is that for supporting none showing tab?

PR #472, opened 2019-11-13
Data-Table Virtual Load Update

daanishnasir:
## Description

- Scrolls to the bottom of the table, make a call to load new data rows to add

- change visibleRows prop to numberOfRowsToLoad

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


angular/src/lib/data-table/data-table-scroll.component.ts
```diff
@@ -240,7 +240,7 @@ export class ScrollComponent implements AfterViewInit, OnDestroy, AfterViewCheck
 
           if (this.scrollLoadingTableViewChild && this.scrollLoadingTableViewChild.nativeElement) {
             this.scrollLoadingTableViewChild.nativeElement.style.display = 'table';
-            this.scrollLoadingTableViewChild.nativeElement.style.top = this.scrollBodyViewChild.nativeElement.scrollTop + 'px';
+            // this.scrollLoadingTableViewChild.nativeElement.style.top = this.scrollBodyViewChild.nativeElement.scrollTop + 'px';
```
KHempel: can we get rid of this commented code here (& line 258) entirely?

angular/src/lib/data-table/examples/data-table-virtual-scroll.component.ts
```diff
@@ -8,13 +8,13 @@ import { Component, OnInit } from '@angular/core';
       [columns]="columns"
       [data]="virtualPeople"
       [scrollable]="true"
-      scrollHeight="250px"
-      [visibleRows]="30"
+      scrollHeight="400px"
+      [visibleRows]="visibleRows"
```
KHempel: is visible rows really necessary?  since it should be all rows currently loaded in the data array whether that's 7 or 50 or 150?

daanishnasir: Little different, I can change the wording of the prop to make it more clear

visibleRows is the next number of table rows to load at a time, you can increment the new tablerows as many as you like. 

This example starts with 20, then loads the additional 20 rows as you scroll while keeping the already loaded rows. When you scroll to the bottom, all rows will be loaded into the table.

KHempel: that still doesn't sound like it fits with what we need lazy load to do

Okay, so we call the API and get the first 50 locations and populate the list with those
then we scroll down, hit the bottom of the list, and a new API call to get the next 50 locations is made
maybe we get back 2 locations, those two locations are added to the list and any further scrolling to the bottom of the list will no longer trigger API calls because we've loaded everything already
but maybe we get back another 50, add those to the list, and keep scrolling until we hit the bottom of the list again and trigger another API call to fetch the next 50 on the list...

the scenario you're describing and what I'm seeing in the example still seem to be dependent on the data being preloaded - basically, no matter how many rows are in the list, all should be visible so the 'visibleRows' would always be data.length

angular/src/lib/data-table/data-table-scroll.component.ts
```diff
@@ -118,6 +120,8 @@ export class ScrollComponent implements AfterViewInit, OnDestroy, AfterViewCheck
   initialized: boolean;
   preventBodyScrollPropagation: boolean;
 
+  page: number = 0;
+
```
KHempel: looks like there's an extra space here that can go away

PR #482, opened 2019-11-19
feat(select): select can filter by multiple fields not just 1, filterValueChange Emitter, custom templates

daanishnasir:
- New Feat: Select Filter can now filter by more than one field 
   -[filterBy]="['name','delete']" ;  filterBy='delete'
- If filterBy prop is not used, will filter by the optionLabel by default
- New Feat: Add filterValueChange emitter when user types 
- New Feat: Add a custom template option to be used for the select dropdown option and a custom template for the selected option so users can provide their own layout to show; added example 
- Add optionClass prop to add css classname to each select option 
- Provide example of how to clear the checkboxes on multi select 

Screen Shot 2019-12-04 at 3 36 43 PM

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


angular/src/lib/select/select.component.ts
```diff
@@ -170,7 +170,7 @@ export class SelectComponent implements AfterContentChecked, ControlValueAccesso
   @Input() optionLabel: string;
   /** @prop show the filter search | false */
   @Input() filter: boolean = false;
-  /** @prop set which key the filter will use | 'label' */
+  /** @prop set which key or keys the filter will use, filtering by multiple keys example: 'name,id' | 'label' */
```
pauljeter: Let's make this a type `string | string[]` so that you can have an array of filters instead of a comma separated string.
Then you pass the array on to the `FilterUtils.filter()` method instead of converting the string into an array.
This will also make the @Input type more clear for the api.

daanishnasir:
Will be adding a new feature on select to this PR

angular/src/lib/select/select.component.ts
daanishnasir: Thanks!
Updated so either works, string or array of strings 
[filterBy]="['name','delete']"
filterBy='delete'

Issue #494, opened 2019-12-02
Angular: md-tooltip does not follow specifications


kjeree:
## Issue Type
Bug report
## Component or Pattern Affected
Angular's `md-tooltip`

## Expected Behavior
The tooltip text should be centered, according to examples at https://momentum.design/components/tooltip/usage

## Current Behavior
The tooltip text is left-aligned.

## Possible Solution
Will create a PR.

## Context
Code does not seem to follow specifications.

## Your Environment

* Library: Angular
* Version used: Latest master
* Browser Name and version: All browsers
* Operating System and version: All operating systems
* Link to your project: N/A

bfbiggs:
Closing as angular library has been forked and removed.

PR #495, opened 2019-12-02
feat(tooltip) Center aligning tooltip text for Angular

kjeree:
## Description
Angular's `md-tooltip` did not follow the specification in https://momentum.design/components/tooltip/style, which says that text alignment should be centered.

## Related Issue
https://github.com/momentum-design/momentum-ui/issues/494

## Motivation and Context
The fix makes the library follow the documented examples. I assume the examples are correct, because other libraries than Angular seem to be consistent with documentation.

## How Has This Been Tested?
After making the (trivial) change, I loaded it up in the Momentum project's Playground, and verified that it now looks good.

## Screenshots:
tooltip_before

**After**:

after_change

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


kjeree:
> Did someone change the design? 

Not to my knowledge, I think the working assumption should be that the Angular component was implemented incorrectly from the get-go. 

> Why do we want all tooltips center aligned at all times? 

Cisco Webex Control Hub has it like that for all code that's using the AngularJS `md-toolkit`, i.e. the overwhelming majority of the product. Now that we've started migrating Control Hub from AngularJS to Angular, we see that Momentum's Angular components behave differently. Because Control Hub has used Momentum for quite some time without anyone complaining, we assume that the AngularJS behavior is correct, and the Angular behavior is incorrect.

>Maybe this should be added as a alternate property?

Perhaps, but I would still recommend defaulting to center aligning.

mathsusan:
Can you kick off a build again to see if the checks will pass?

daanishnasir:
@kjeree can you take out the 🐛 in the committ message and re-format

feat(component-name) : center aligning tooltip text

kjeree:
> @kjeree can you take out the 🐛 in the committ message and re-format

Sure. I by the way used `yarn commit` to create the original commit; if this tool produces commit messages on a format we don't want, perhaps we should update (or remove) it?

pauljeter:
@kjeree that's a valid point. I'll look at it.

PR #505, opened 2019-12-17
fix(select): dynamic filter height fix

daanishnasir:
- fix console errors
- list will now shrink dynamically, no fixed height 
- no horizontal scroll bar 

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/angular/src/lib/select/select.component.ts
+++ b/angular/src/lib/select/select.component.ts
@@ -80,7 +80,6 @@ export class SelectChange {
       [cdkConnectedOverlayOrigin]="trigger"
       [cdkConnectedOverlayOpen]="overlayOpen"
       [cdkConnectedOverlayWidth]="this.anchorWidth"
-      [cdkConnectedOverlayHeight]="this.options.length > 8 ? this.scrollHeight : 'auto'"
       [cdkConnectedOverlayOffsetY]="6"
       [cdkConnectedOverlayPanelClass]="['md-select__dropdown', overlayClass]"
       [cdkConnectedOverlayHasBackdrop]="true"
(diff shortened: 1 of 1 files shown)


angular/src/lib/select/select.component.ts
```diff
@@ -115,10 +114,9 @@ export class SelectChange {
 
       <div
         class="md-list md-list--vertical"
-        [style.height]="this.options.length > 8 ? scrollHeight : 'auto'"
-        [style.max-height]="scrollHeight"
+        [style.max-height]="this.options && this.options.length > 8 ? scrollHeight : '350px'"
         role="listbox"
-        style="align-items: normal; overflow: scroll;"
+        style="align-items: normal; overflow: scroll; overflow-x:hidden"
```
ArthusLiang: Can we move this to core?

daanishnasir:
PROVISION-5742

angular/src/lib/select/select.component.ts
```diff
@@ -115,10 +114,9 @@ export class SelectChange {
 
       <div
         class="md-list md-list--vertical"
-        [style.height]="this.options.length > 8 ? scrollHeight : 'auto'"
-        [style.max-height]="scrollHeight"
+        [style.max-height]="this.options && this.options.length > 8 ? scrollHeight : '350px'"
```
mathsusan: 350px?  where did that come from?

daanishnasir: A default height was needed to avoid a large list showing

rovshenn: Can we use rem instead of px to make sure it is a relative measure?

Issue #507, opened 2019-12-17
(select) - change events triggered on arrow key navigation of open select list


KHempel:
When the md-select list is open, arrow key navigation of the select causes the change events to be triggered each time, even though the selected option has not actually been changed yet; can potentially cause breaking behavior with keyboard navigation, as the change event might not only be called multiple times, but ultimately the select may end up being closed without actually making any changes

## Issue Type
Bug - accessibility issue

## Component or Pattern Affected
Select

## Expected Behavior
Arrow navigation of an open list should not call the change event.  The change event should only be called when an item on the list is selected and the list closes.

PR #510, opened 2019-12-19
fix(select): onChange value emitter only fires with enter and click i…

daanishnasir:
…nstead of arrow nav

PROVISION-5762

(Angular Toolkit - md-select) Arrow navigation of the open list triggers change events - Accessibility issue/Minor breaking issue

Because the change event is being fired when the selection hasn't actually changed, this can cause inconsistent/bad behavior as changes occur behind the scenes based on options that are not actually selected
this is specifically an accessibility issue because it affects keyboard navigation

## Description

- value change emitter only fired on click of select item or on 'Enter';
- Placeholder value only changes if selected item is actually selected from click or enter

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/angular/src/lib/select/examples/select-custom-content.component.ts
+++ b/angular/src/lib/select/examples/select-custom-content.component.ts
@@ -24,10 +24,9 @@ import { Component, OnInit } from '@angular/core';
           </div>
         </ng-template>
 
-
         <!-- If you want to provide a custom selected option label -->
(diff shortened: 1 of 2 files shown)


daanishnasir:
#507

PR #519, opened 2020-01-16
fix(select): writeValue working for formControl

JohnChambersBestFriend:
## Description
Used with a formControl group, patchValue and setValue were not reflecting new values in the UI.  The root cause was using the `===` operator to compare objects (always returns false).  This was replaced with the lodash `isEqual` method which is functional when comparing objects.  The change was to `writeValue` and the underlying `find/selectUpdatedItem` methods.  I consolidated the functionality for pressing `enter` with an item selected, with that of `writeValue` as well.

Additionally, I added the select-forms.component to the kitchen sink as a route and added some additional controls that could be used to manually test the component in action.

Lastly, when I ran the project it install, it updated the yarn.lock file, so that's in there as well.

## Screenshots:
**After**:
Screen Shot 2020-01-16 at 3 18 14 PM

## Types of changes

- [ ] Documentation update
- [x] Build related changes (yarn.lock updated)
- [ ] CI related changes

## Checklist:

### DIFF
--- a/angular/src/lib/select/select.component.ts
+++ b/angular/src/lib/select/select.component.ts
@@ -7,7 +7,9 @@ import { Subscription } from 'rxjs';
 import { ButtonComponent } from '../button/button.component';
 import { TableService } from '../data-table/data-table.service';
 import { TemplateNameDirective } from '../data-table/shared';
-import { isEqual } from 'lodash';
+import { FilterUtils } from '../utils/filterUtils/filters';
+import { SelectService } from './select.service';
+
 
 const SELECT_VALUE_ACCESSOR: any = {
   provide: NG_VALUE_ACCESSOR,
@@ -472,7 +474,7 @@ export class SelectComponent implements AfterContentChecked, AfterContentInit, C
 
       if (this.isMulti) {
         this.toggleRowWithCheckbox(this.selectedOption.value);
-      } else {
+      } else if (emit) {
         this.close();
       }
     }
--- a/angular/src/lib/select/select.component.ts
+++ b/angular/src/lib/select/select.component.ts
@@ -1,29 +1,10 @@
-import {
-  AfterContentChecked,
-  AfterContentInit,
-  ChangeDetectorRef,
-  Component,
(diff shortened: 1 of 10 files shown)


angular/src/lib/select/select.component.ts
```diff
@@ -489,7 +456,29 @@ export class SelectComponent implements AfterContentChecked, AfterContentInit, C
     }, this.closeDuration);
   }
 
-  selectItem(event, option) {
+  private forceSelectItem(emit: boolean = false) {
```
JohnChambersBestFriend: So, this is a cut-and-paste of the code used for handling the enter key (I wanted writeValue to share the same flow).  However, when coming in from writeValue, `emit` will be false so there will be no `onChange` event emitted.

angular/src/lib/select/select.component.ts
```diff
@@ -489,7 +456,29 @@ export class SelectComponent implements AfterContentChecked, AfterContentInit, C
...
+      }
+
+      if (this.tableRowIndex >= 0) {
+        this.tableService.onSelectChange(this.value, this.tableRowIndex);
+      }
+
+      if (this.isMulti) {
+        this.toggleRowWithCheckbox(this.selectedOption.value);
+      } else if (emit) {
+        this.close();
```
JohnChambersBestFriend: This close was tripping up the last unit test case which tests clicking the button on the select component and then checking if the overlay is visible.  It wasn't happening because the `click` method under the covers calls `setValue` which then comes in here via `writeValue` which subsequently closes the overlay.  After looking at this, I think we want both the `onSelChange` and the `toggleRowWithCheckBox` to be called because they are setting component state, we just don't want the action of `close` in the case of single-select.

Where this could crop up is sort of a race condition where the user clicks the select at the same time that we are updating it via patchValue or setValue.  In this case, I believe the select component will reflect the change in the value of the control but not close the overlay unexpectedly on the user.

bfbiggs:
Can you resolve the conflicts?  Needed to merge the other select PR first.

Issue #503, opened 2019-12-13
(select): Focus lost on selection from list (accessibility issue)


KHempel:
## Issue Type 
Bug Report - Accessibility issue

## Expected Behavior
When an item is selected from the list and the select is closed, focus should return to the trigger element 

## Current Behavior
Focus is lost and subsequent tab through places focus back at the top of the place, interrupting keyboard accessibility flows

## Possible Solution
on list close, focus should be auto-returned to the trigger element

PR #521, opened 2020-01-21
fix(select): set scroll to active item when overlay is opened

daanishnasir:
- Set scroll to active item when select overlay is open

- Able to pass in ngModel value to appear on page load selected placeholder

- Active item background color will go back to the selected Value if backdrop clicked

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/angular/src/lib/select/select.component.ts
+++ b/angular/src/lib/select/select.component.ts
@@ -26,6 +26,7 @@ import { NG_VALUE_ACCESSOR, ControlValueAccessor } from '@angular/forms';
 import { SelectService } from './select.service';
 import { TableService } from '../data-table/data-table.service';
 import { TemplateNameDirective } from '../data-table/shared';
+import { isEqual } from 'lodash';
 
 const SELECT_VALUE_ACCESSOR: any = {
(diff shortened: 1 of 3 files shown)


daanishnasir:
#503

angular/src/lib/select/select.component.ts
```diff
@@ -611,12 +613,18 @@ export class SelectComponent implements AfterContentChecked, AfterContentInit, C
     if (this.filterViewChild && this.filterViewChild.nativeElement) {
       this.filterViewChild.nativeElement.focus();
     }
+    const activeItem = document.querySelector('.md-select-item--focus');
```
ArthusLiang: Do we need to support the browser which does not support querySelector?

daanishnasir: Looks to be supported by all browsers appearing here; https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector

Issue #537, opened 2020-02-05
[Modal] Modal gets closed when pressing any key while focus is on the close button


d-silvas:
We have found 2 problems, regarding accessibility, with the modal closing button:
1. The button doesn't show any box shadow when the focus is on it.
1. If we tab until the focus is on the button and then press *any* key on the keyboard, the modal will close. This is due to the `(keydown)` event listener on the button.

## Issue Type
Bug Report

## Component or Pattern Affected

`ModalHeaderComponent`, in the Angular library.

## Expected Behavior

1. We expect to see a blue box shadow when the focus is on the modal close button (like with any other button)
![expected](https://user-images.githubusercontent.com/15788150/73864807-27210580-483a-11ea-97e5-6af2db94ba7d.png)
1. The modal should only close when pressing Enter or Space on the close button (this would cause the browser to trigger a click event).

## Current Behavior

1. No styling shown when the button is focused
![current](https://user-images.githubusercontent.com/15788150/73864784-1f616100-483a-11ea-9f63-8b725ccb8fa8.png)
2. The modal gets closed when pressing any key on the button.

## Possible Solution

PR provided

## Steps to Reproduce (for bugs)

1. Open a modal.
2. Tab until the focus is on the closing button. See that there is no style applied.
3. Press any key. The modal will close.

## Your Environment

* Library: Angular
* Version used: 10.5.0
* Browser Name and version: Chrome and Firefox, latest versions
* Operating System and version: Windows 10

PR #538, opened 2020-02-05
fix(modal): remove keydown event on close button

d-silvas:
## Description

1. Show the box shadow when the modal close button is focused.
2. Remove keydown event.

Note that, when pressing Enter or Space while the focus is on the button, the browser triggers a click event and the modal will get closed.

## Related Issue

https://github.com/momentum-design/momentum-ui/issues/537

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/angular/src/lib/modal/modal-header.component.ts
+++ b/angular/src/lib/modal/modal-header.component.ts
@@ -20,7 +20,6 @@ import { ModalRef } from './modal-ref';
     *ngIf="showCloseButton"
     class="md-close md-modal__close"
     (click)="close()"
-    (keydown)="handleKeydown($event)"
     aria-label="close"
   >
   </button>
--- a/core/scss/components/modal/modal.scss
+++ b/core/scss/components/modal/modal.scss
@@ -43,10 +43,6 @@
           @include flex;
 
           border: none;
-
-          &:focus {
-            box-shadow: none;
-          }
         }
       }
 


angular/src/lib/modal/modal-header.component.ts
```diff
@@ -20,7 +20,6 @@ import { ModalRef } from './modal-ref';
     *ngIf="showCloseButton"
     class="md-close md-modal__close"
     (click)="close()"
-    (keydown)="handleKeydown($event)"
```
mathsusan: so .. can you still use the keyboard to close the modal?

d-silvas: Yes, the browser automatically triggers a click event when you press the Enter or Space keys. This will call `close()` and close the modal.

mohammedzamakhan: essentially this block of code will make sure that `ESC` key will close the modal https://github.com/momentum-design/momentum-ui/blob/ab519026a8310b1452132eac1bdfbbedff8f81d8/angular/src/lib/modal/modal-header.component.ts#L45

d-silvas: @mohammedzamakhan That's correct. We don't have any issue with that though. The listener is on `document`, so the ESC key will always close the modal.

The `(keydown)` event **on** the button is problematic because, when the focus is on the button, pressing *any* key causes the modal to be closed. For example: if you open a modal and just press Tab repeatedly, you will eventually close the modal (the focus will eventually be on the button, and then, one more Tab key press will close the modal).

PR #542, opened 2020-02-13
feat(angularjs): allow 'mdTokenField' to work with UpgradeComponent

brspence:
## Description

- replace preLink/postLink with single link function
- remove `replace` and select correct element for tokenfield jquery

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/angularjs/src/lib/components/tokenfield/tokenfield.component.ts
+++ b/angularjs/src/lib/components/tokenfield/tokenfield.component.ts
@@ -9,7 +9,9 @@ export interface ITokenField extends ng.IScope {
 export class TokenField implements ng.IDirective {
   constructor(private $timeout: ng.ITimeoutService) {}
 
-  public preLink: ng.IDirectiveLinkFn = (scope: ITokenField, elem: any, attr: ng.IAttributes) => {
+  public link: ng.IDirectiveLinkFn = (scope: ITokenField, elem: any, attr: ng.IAttributes) => {
+    const tokenfieldElem = elem.find('> input');
+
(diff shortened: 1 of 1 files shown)


angularjs/src/lib/components/tokenfield/tokenfield.component.ts
```diff
@@ -76,18 +78,16 @@ export class TokenField implements ng.IDirective {
...
-        tokenField.addClass('form-control-static').removeClass('form-control');
-        let tokenInput = tokenField.find('input').last();
+        let tokenFieldElemParent = tokenfieldElem.parent();
+        tokenFieldElemParent.addClass('form-control-static').removeClass('form-control');
+        let tokenInput = tokenFieldElemParent.find('input').last();
         tokenInput.attr('readonly', 'true');
 
         // Make the input editable when token is being edited
-        elem
+        tokenfieldElem
```
divide-by-zero-443: [Question] if the tokenfieldElem is changed here when removing readonly, is there any place else where this is added back (on a user action, fox example). If so, shouldn't tokenfieldElem be changed even there ?

brspence: Mmm nope because the static setting is just an initial input value and not expression evaluated to change the component behavior when the input changes. Not modifying any other functionality here accept modifying the element references now that they're children of this component without using the `replace` option.

PR #565, opened 2020-03-17
feat(List): new prop for enabling event propagation

Akashdeep-Sarin:
Lists normally do stop event bubbling by `e.stopPropagation`. This becomes an issues sometimes, when the key-down events are needed for maintaining focus / or accessibility.

## Description

Adding a new props `propogateKeyDownEvents` for enabling event propagation.
Also adding sample in React KitchenSink

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


react/src/lib/List/examples/KitchenSink.js
```diff
@@ -1,17 +1,34 @@
...
 import React from 'react';
 import {
   ListDefault,
   ListListItemHeader,
   ListListItemSeparator,
+  CheckBoxForKeyDownEvents,
 } from './index';
 
 export default class ListKitchenSink extends React.Component {
+  state ={
```
bfbiggs: Can we move this to the individual example instead of the KitchenSink.

react/src/lib/List/index.js
```diff
@@ -422,6 +423,8 @@ List.propTypes = {
   itemRole: PropTypes.string,
   /** @prop Callback function invoked by user selecting an interactive item within List | null */
   onSelect: PropTypes.func,
+  /** @prop Disables the stopPropagation() & preventDefault() for ArrowKey Events | false */  
+  propogateKeyDownEvents: PropTypes.bool,
```
bfbiggs: Let's call this shouldPropagateKeyDown...to match the other booleans of this component.

react/src/lib/List/examples/Default.js
```diff
@@ -20,3 +23,12 @@ export default class ListItemDefault extends React.PureComponent {
     );
   }
 }
+
+ListItemDefault.propTypes = {
```
bfbiggs: Default should not have anything added to it.

react/src/lib/List/examples/CheckBoxForKeyDownEvents.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Checkbox } from '@momentum-ui/react';
+
+export default class CheckBoxForKeyDownEvents extends React.PureComponent {
```
bfbiggs: This should contain both state and any necessary components to show the example (i.e. List)

react/src/lib/List/examples/KitchenSink.js
Akashdeep-Sarin: Moved to new component,

react/src/lib/List/index.js
Akashdeep-Sarin: Done.

react/src/lib/List/examples/Default.js
Akashdeep-Sarin: Reverting `Default` to old state.

react/src/lib/List/examples/CheckBoxForKeyDownEvents.js
Akashdeep-Sarin: Moved List, state and checkbox to this component. 
Renamed it to `ListWithEventBubbling`

react/src/lib/List/index.js
```diff
@@ -177,14 +177,14 @@ class List extends React.Component {
       case 37:
         this.getNextFocusedChild(items, tgt, -1);
         this._needsRefocus = true;
-        flag = true;
+        if(!shouldPropagateKeyDown) flag = true;
```
andrewholsted: Is this backwards compatible? It's hard for me to tell from the logic/case statement. We should make the default whatever it was and have the new prop change that behavior. If that is how it works then I'll approve

yjlo: +1

Akashdeep-Sarin: @andrewholsted  - Yes, it is backwards compatible. 
[here](https://github.com/momentum-design/momentum-ui/blob/f23e9e2f6af5a7fbb785fad340f2768f6d308966/react/src/lib/List/index.js#L457), I have set the default props to `false` and in case of default value `false`, [line](https://github.com/momentum-design/momentum-ui/blob/f23e9e2f6af5a7fbb785fad340f2768f6d308966/react/src/lib/List/index.js#L187) and [line](https://github.com/momentum-design/momentum-ui/blob/f23e9e2f6af5a7fbb785fad340f2768f6d308966/react/src/lib/List/index.js#L180) would work  as previously. 
So, whenever we don't have the `shouldPropagateKeyDown` props, it should work as before.

react/src/lib/List/examples/KitchenSink.js
```diff
@@ -3,15 +3,17 @@ import {
...
 } from './index';
 
 export default class ListKitchenSink extends React.Component {
   render() {
     return (
       <React.Fragment>
-        <ListDefault />
-        <ListListItemHeader />
-        <ListListItemSeparator />
+          <ListDefault/>
```
bfbiggs: Can you check your tabing/spaces compared to the current library?

react/src/lib/List/examples/ListWithEventBubbling.js
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable no-console */
+/* eslint-disable jsx-a11y/no-static-element-interactions */
+import React from 'react';
+import {Checkbox, List, ListItem} from '@momentum-ui/react';
+
+export default class ListWithEventBubbling extends React.PureComponent {
+
+    state = {
```
bfbiggs: Could be handled with hook ;);

react/src/lib/List/index.js
```diff
@@ -177,14 +177,14 @@ class List extends React.Component {
...
         break;
 
       case 39:
       case 40:
         this.getNextFocusedChild(items, tgt, 1);
         this._needsRefocus = true;
-        flag = true;
+        if(!shouldPropagateKeyDown) flag = true;
```
bfbiggs: This is backwards compatible as long as no overrides come in the future...I'm not certain this is the way to handle it but it's difficult to know until we have all other possible edge cases @andrewholsted

bfbiggs: Just want to mention: One of the challenges of including accessibility in the library.

react/src/lib/List/examples/KitchenSink.js
Akashdeep-Sarin: Done.

react/src/lib/List/index.js
Akashdeep-Sarin: Yeah.. Most of Accessibility changes find their way into changes in momentum-ui code. 😃

Issue #575, opened 2020-03-26
Compile CSS Modules from SASS Modules


futurethang:
I am in need of the library to provide all of the components CSS available as individual modules for importing into a Web Components extension of the Momentum UI kit.

## Issue Type

Feature Request 

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Library: 
* Version used: 
* Browser Name and version: 
* Operating System and version: 
* Link to your project:

futurethang:
Solved by #576

PR #576, opened 2020-03-26
feat(core): add component CSS modules to build process

futurethang:
Add package.json script to include SASS compile to build, using the compileCSSModules.js tool. Modified compileCSSModule.js tool to include proper fs functions and promise callback, as well as refined directory structure

## Description
Changes to `package.json` and `compileCSSModules.js`

## Related Issue
#575 

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


core/tools/compileCSSModules.js
```diff
@@ -64,8 +60,9 @@ const modules = [
 
 const compileCSSModules = () => {
   modules.forEach((module, idx) => {
-    const outFile = path.resolve(__dirname, `../css/components/${module}.css`);
-    const mapFile = path.resolve(__dirname, `../css/components/${module}.map.css`);
+    const componentsDirectory = path.resolve(__dirname, `../css/components/${module}`);
```
ArthusLiang: @bfbiggs will this affect other things? The structure changes here

core/tools/compileCSSModules.js
```diff
@@ -26,11 +26,7 @@ const modules = [
   'date-picker',
   'drawer',
   'date-range-picker',
-  'editable-textfield',
   'overlay',
-  'footer',
-  'hero',
-  'icon',
```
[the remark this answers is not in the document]
futurethang: Those component folders do not have their own module.scss files to compile.

core/tools/compileCSSModules.js
futurethang: Correct, it enables the components to each have their own folders. Previously it was resolving all CSS  and map files inside one components directory, less organized.

core/tools/compileCSSModules.js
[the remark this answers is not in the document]
bfbiggs: I think that's a temporary issue in terms of these components.

core/tools/compileCSSModules.js
bfbiggs: Correct but you need to verify with @pauljeter if this affects what is output on the cdn.

core/tools/compileCSSModules.js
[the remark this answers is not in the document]
futurethang: I see. As a remedy, would it be better to have that array list generated by searching over the `css/components` directory for only components with the module.scss, or better to keep the list there comprehensive but error handle when module.scss is missing?

core/tools/compileCSSModules.js
futurethang: Copy that. I could use either solution ultimately, but thought this was a logical structure. What do you think @pauljeter ?

pauljeter: Not sure how it would be better organized? Instead of flat component css files, there would be a folder for each component with only 1 file. I'm good either way. The path will be a little redundant with the folder structure. 
`css/components/button/button.css` vs `css/components/button.css

core/tools/compileCSSModules.js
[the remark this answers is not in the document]
pauljeter: I agree that we should be searching for all of the module.scss files instead of using a defined list. If anything, the list should be the files to ignore.

core/tools/compileCSSModules.js
pauljeter: Also, @bfbiggs, we are not currently publishing the components css to NPM or the CDN. This code was here from when I was exploring the option of publishing the modules.

core/tools/compileCSSModules.js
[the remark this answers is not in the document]
futurethang: Per discussion with @pauljeter, We'll move forward with the list that is here, and I will make a new issue with notes on my efforts so far and come back to improve in another PR.

core/tools/compileCSSModules.js
futurethang: I'm re-working this bit to skip the map files and compile to a flat `css/components` directory

PR #601, opened 2020-04-28
feat(Listitem): focus on inner children only when current listitem is in focus

yjlo:
## Description

I added 2 props `tabbableChildrenQuery` and `focusLockTabbableChildren` so that we can choose to do this for listitems that have intractable elements inside. Without this, the user will have to tab through all avatars for example.
**NOTE: I tried using focus lock but that messed everything up**

## Related Issue

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-126886

## How Has This Been Tested?

I manually imported it into web.webex.com and tested it with one of our components

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


aimeex3:
does this take into account disabled items? will it skip over disabled items?
i guess if you pass the query to not select the disabled item, it won't tab through them

yjlo:
@aimeex3, ya I was thinking it would be better to give people the choice

react/src/lib/ListItem/index.js
```diff
@@ -70,19 +70,79 @@ class ListItem extends React.Component {
...
     parentKeyDown && parentKeyDown(e, { value, label, eventKey });
   }
 
+  handleBlur = (e, focusLockTabbableChildren, tabbableChildrenQuery) => {
+    const { onBlur } = this.props;
+
+    if (focusLockTabbableChildren) {
+      if (e.target) {
+        const currListItem = e.target.closest('.md-list-item');
+        const tabbableChildren = currListItem.querySelectorAll(tabbableChildrenQuery);
```
andrewholsted, line 131: If someone uses list item outside of a list, I think this will throw an error since `currListItem` will be undefined

react/src/lib/ListItem/index.js
```diff
@@ -70,19 +70,79 @@ class ListItem extends React.Component {
...
+        }
+      }
+    }
 
     e.persist();
     onKeyDown && onKeyDown(e);
     parentKeyDown && parentKeyDown(e, { value, label, eventKey });
   }
 
+  handleBlur = (e, focusLockTabbableChildren, tabbableChildrenQuery) => {
```
andrewholsted, line 125: Do you have a sense for how often this blur event would get fired? I recall @aimeex3 having events trigger on multiple levels for file drag/drop. I want to make sure the same doesn't apply here.

react/src/lib/ListItem/index.js
andrewholsted: Probably need a check for tabbableChildren too? Is it possible that someone passes the prop but not the query?

react/src/lib/ListItem/index.js
yjlo: It shouldn't be too bad since they have to have focusLockTabbableChildren true for it to do anything else.

react/src/lib/ListItem/index.js
yjlo: Forgot about that adding checks now

react/src/lib/ListItem/index.js
yjlo: As for when you do use the prop, I can try to limit it from changing the tabindex too much

react/src/lib/ListItem/index.js
```diff
@@ -70,19 +70,84 @@ class ListItem extends React.Component {
...
-  handleKeyDown = (e, eventKey) => {
+  changeTabIndex = (tabbableChildren, index) => {
+    for (let i = 0; i < tabbableChildren.length; i++) {
+      if (tabbableChildren[i].tabIndex === (index === 0 ? -1 : 0)) {
+        tabbableChildren[i].tabIndex = index;
+      }
+    }
+  }
+
+  handleKeyDown = (e, eventKey, focusLockTabbableChildren, tabbableChildrenQuery) => {
```
bfbiggs, line 81: Getting to the point where it would make sense to just have  an opts object passed in.

react/src/lib/ListItem/index.js
```diff
@@ -250,6 +321,8 @@ ListItem.propTypes = {
   label: PropTypes.string,
   /** @prop external link associated input | '' */
   link: PropTypes.string,
+  /** @prop Callback function invoked by user changing focus from current ListItem ListItem | null */
```
bfbiggs, line 324: Additional ListItem here.

react/src/lib/ListItem/index.js
yjlo: I tried, to, but it was failing builds when I use .shape

react/src/lib/ListItem/index.js
yjlo: Thanks :)

Issue #604, opened 2020-05-06
[Modal] Angular modal shows underneath the backdrop


d-silvas:
UPDATE: for anyone encountering this issue, the solution is to import Momentum's custom overlay CSS:
```
@import '~@momentum-ui/core/scss/components/cdk-overlay/cdk-overlay';
```
instead of Angular's:
```
@import '~@angular/cdk/overlay-prebuilt.css';
```

----

## Expected Behavior

![after](https://user-images.githubusercontent.com/15788150/81172467-3416d780-8f96-11ea-8800-fb5a74260f31.PNG)

## Current Behavior

![before](https://user-images.githubusercontent.com/15788150/81172434-26615200-8f96-11ea-9b4f-6ce10ca6ce12.PNG)

## Steps to Reproduce (for bugs)

Use the project's Angular playground with any of the modal example components, like, for instance:
```
<example-modal-default></example-modal-default>
```

## Your Environment

* Library: Angular
* Version used: 11.7.3
* Browser Name and version: Chrome and Firefox latest versions
* Operating System and version: Windows 10
* Link to your project:

PR #605, opened 2020-05-06
fix(modal): fix angular modal showing underneath the backdrop

d-silvas:
## Description

The modal backdrop is currently showing over the modal. We can see what the issue is by examining the code (in the following picture, the fix has already been applied):

html

In summary:
- Angular creates a `div.cdk-overlay-backdrop` element, which we add the `.md-modal__backdrop` class to. Momentum's class sets the z-index of this element to **1040**.
- Angular also creates a `div.cdk-global-overlay-wrapper` element as a *sibling* of the previous one, and sets its z-index to **1000**.
- As a consequence, any child under `div.cdk-global-overlay-wrapper` will be shown underneath `div.cdk-overlay-backdrop`.
- Momentum's `.md-modal` element actually has its z-index set to 1050, but it will always show underneath `div.cdk-global-overlay-wrapper`.
----

The provided changes fix this problem by setting `.md-modal__backdrop` z-index to 1000, effectively evening out the z-indexes of the 2 components created by Angular.

I appreciate this solution might not be the best because we are modifying global styles from within the modal Component. Please let me know any suggestion or better ways of fixing the problem. Other things I've considered are:
- Modifying the `.md-modal__backdrop` CSS directly. However this can affect other libraries (angularjs, react) in ways I can't foresee.
- Adding a set of utility classes to modify the z-index of an element. Similar to the current margin and padding classes.

## Related Issue

https://github.com/momentum-design/momentum-ui/issues/604

## Screenshots:
**Before** (If applicable):

before

**After**:

after

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/angular/src/lib/modal/modal.component.ts
+++ b/angular/src/lib/modal/modal.component.ts
@@ -39,7 +39,11 @@ export type SizeType =
   .md-modal.md-modal--full {
       position: static;
     }
-  `]
+  `, `
+  ::ng-deep .md-modal__backdrop {
+      z-index: 1000;
+    }
+  `],
 })
 export class ModalComponent implements OnInit, AfterViewInit {
   /** @prop Determines the visibility and ability to edit the backdrop of the Modal | true */


angular/src/lib/modal/modal.component.ts
```diff
@@ -39,7 +39,11 @@ export type SizeType =
   .md-modal.md-modal--full {
       position: static;
     }
-  `]
+  `, `
+  ::ng-deep .md-modal__backdrop {
+      z-index: 1000;
+    }
+  `],
```
mathsusan, line 46: this really doesn't seem like what we should do.

mathsusan: In core, we included https://github.com/momentum-design/momentum-ui/tree/master/core/scss/components/cdk-overlay and adjusted the z-index .. so if you are importing the cdk.css you should not.

d-silvas: Cool thanks, will use those styles and close this PR as it is a non-issue.

PR #610, opened 2020-05-18
Remove libraries

pauljeter:
This PR removes Angular, AngularJS and Vue libraries from the momentum-ui mono-repo. These libraries will exist in their own repos and will not be maintained with the core and react libraries.

[angular](https://github.com/momentum-design/momentum-angular)
[angularjs](https://github.com/momentum-design/momentum-angularjs)
[vuejs](https://github.com/momentum-design/momentum-vue)


andrewholsted:
HERE WE GOOOOOOO

PR #618, opened 2020-06-19
 feat(list): can specify navigation type

yjlo:
Added a prop to List
## Description

Added `navigationDirection` prop so that when navigating through the list you can choose to use UP/DOWN, LEFT/RIGHT, or both.
## Motivation and Context

Visually a horizontal list, you may not want UP/DOWN to navigate through the list. Same ver vertical list and LEFT/RIGHT

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


react/src/lib/List/index.js
```diff
@@ -426,6 +443,8 @@ List.propTypes = {
   id: PropTypes.string,
   /** @prop Optional tabType prop type to manually set child role | 'listitem' */
   itemRole: PropTypes.string,
+  /** @prop Restricts the traversal of the list with either UP/DOWN, LEFT/RIGHT, or both | 'both' */
+  navigationDirection: PropTypes.oneOf(['vertical', 'horizontal', 'both']),
```
bsiboliban-csco, line 447: Is `both` a common value that we use? Seems like `any` or `all` might be a better choice.

react/src/lib/List/index.js
```diff
@@ -174,29 +174,45 @@ class List extends React.Component {
         flag = true;
         break;
 
-      case 38:
-      case 37:
-        this.getNextFocusedChild(items, tgt, -1);
-        this._needsRefocus = true;
-        if(!shouldPropagateKeyDown) flag = true;
+      case 38: // UP
+        if (navigationDirection === 'both' || navigationDirection === 'vertical') {
```
bsiboliban-csco, line 178: Might consider assigning `navigationDirection === 'both' || navigationDirection === 'vertical'` to a variable to make clear the intent. `isVerticalOrAny`? Same goes for when you're checking for `horizontal`.

react/src/lib/List/index.js
yjlo: I'm not really sure what we use, I feel like any is too vague and all as well.

react/src/lib/List/index.js
yjlo: Fair will make appropriate changes.

PR #622, opened 2020-06-25
fix(popover): clear popover delayCheckHover timeout

robstax:
## Description
delayCheckHover creates a timeout that we aren't keeping track of, which is a memory leak. this PR fixes that. everything else is formatting noise.

## Motivation and Context
Memory leak and address a warning from React when we setState on an unmounted component.

## How Has This Been Tested?
it's difficult to consistently repro the issue, but logically this fixes it.

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


react/src/lib/Popover/index.js
```diff
@@ -190,13 +187,16 @@ class Popover extends React.Component {
       switch (e.which) {
         case 9:
           if (tabbableElements.length) {
-            if (this.props.closeOnFocusLeavesContent) { // if closeOnFocusLeavesContent = true
-              if (e.shiftKey) { // SHIFT + TAB
+            if (this.props.closeOnFocusLeavesContent) {
+              // if closeOnFocusLeavesContent = true
```
aimeex3, line 191: don't think we need this comment. seems to repeat the code exactly :)

robstax: i didn't write the comment. formatting the file moved it to the next line :P

PR #629, opened 2020-07-15
feat(popover): fix copy paste from contact card bug

dalipkumar703:
This PR add changes in Popover handleClose to allow copy from popover.
## Related Issue

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-158311

## Motivation and Context

In cantina we are unable to copy mailid from contact card that display on popover hover.
## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


react/src/lib/Popover/index.js
```diff
@@ -80,8 +80,11 @@ class Popover extends React.Component {
 
   handleClose = e => {
     const { onClose } = this.props;
-
-    this.setState({ isOpen: false }, onClose && onClose(e));
+    const {keyCode} = e;
+    //allow to copy text on popover
+    if (!(keyCode === 17 || keyCode === 91 || keyCode === 67)) {
```
holsted, line 85: I think it also autocloses whenever I try to highlight and copy/paste. Can we do anything about that (like not blur/close if the mouse is over the contact card?)

dalipkumar703: I tried to highlight using double click and copy text, it's working fine.

yjlo: I can confirm, You can copy with mouse, but you have to make sure you aren't making the mouse go past the popover or else it'll close before you can copy. https://stackoverflow.com/questions/4712310/javascript-how-to-detect-if-a-word-is-highlighted
Seems like we maybe able to detect if something is being highlighted and we can prevent the contact card from closing. Would need a defined way to get out of the highlight. it'll get tricky if we try to use detect highlight of text since it can be highlighted but not copied and we never close the popover etc...

yjlo: A long term easy solution would be to have a copy button, we also need this for meeting passwords too. Emails also get cut off anyway from being too long.

yjlo: Screen Shot 2020-07-17 at 11 39 01 AM

dalipkumar703: @holsted what you suggest here? Should we create another button on popup for copying or to find way to not get it close if there is highlight text and mouse go out of popover. JFYI, copying from popover working whenever mouse is over the popover with these changes.

yjlo: @dalipkumar703 I think it's time we get a design spec made for it and implement it :)
Something similar to this
Screen Shot 2020-07-20 at 2 10 11 PM

dalipkumar703: Okay, I will DM to know more detail

yjlo: Consult Andrew first though

holsted: I like @yjlo recommendation. See if we can get something similar to github from Alex Rosandick? I think there is a new design. In the mean time, we can merge this if it allows copy/paste

Issue #431, opened 2019-10-18
Avatar with prop `type="bot"` and an `onClick` prop has broken styles.


robstax:
## Expected Behavior
If I pass an `onClick` prop to an Avatar with `type="bot"`, the `:after` pseudo-element should not take up the entire avatar.

## Current Behavior
If I pass an `onClick` prop to an Avatar with `type="bot"`, the `:after` pseudo-element takes up the entire height and width of the Avatar.

## Steps to Reproduce (for bugs)
Go to https://momentum.design/components/avatar/code and scroll down the "Bot" example.
Use React dev tools to edit the `onClick` prop (set it to true so the `md-avatar--clickable` class gets applied.

## Your Environment

* Library: React
* Version used: 23.0.5
* Browser Name and version: Chrome 78
* Operating System and version: macOS 10.14.6

PR #634, opened 2020-07-30
fix: fixed prob with avatar type bot and click

ionutrusu-bytex:
## Description

https://github.com/momentum-design/momentum-ui/issues/431
fixed prb with avatar type both and onclick set to true

## Screenshots:
**Before** (If applicable):

**After**:

image

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


bfbiggs:
Were you able to look at the styles to see why this is happening?

react/src/lib/Avatar/index.js
```diff
@@ -163,7 +163,7 @@ class Avatar extends React.Component {
       <div
         className={
           'md-avatar' +
-          `${(onClick && ` md-avatar--clickable`) || ''}` +
+          `${((onClick && type!=='bot') && ` md-avatar--clickable`) || ''}` +
```
bfbiggs, line 166: If the style isn't fixable then let's remove the extra parens here and add spaces around the '!=='

ionutrusu-bytex: @bfbiggs fixed

ionutrusu-bytex:
@holsted 
image bots are still clickable, obs how i pass onClick prop true and type bot...on click u can see the alert window

bfbiggs:
> Sorry, why are bots not clickable?

@holsted yeah I took a look - looks like both bot style and clickable style both modify the :after pseudo element. As long as the lack of the clickable styles aren't required by design then I think this fix is sufficient.  My question is instead of this patch - should we actually look to move bot to function the same way as notification badge and failure badge. This is the only type attribute that modifies the :after element like this.

`     {failureBadge && <span className='md-avatar__failure-badge' />}  
{hasNotification && <span className='md-avatar__notification-badge' />}`

aimeex3:
> My question is instead of this patch - should we actually look to move bot to function the same way as notification badge and failure badge. This is the only type attribute that modifies the :after element like this.

agree with @bfbiggs. that way the onclick behavior still works correctly for bot avatars

ionutrusu-bytex:
@bfbiggs is this PR still valid?

PR #703, opened 2020-11-24
fix(components): fixed close/open modal function - CAX-829

helen-v:
## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/web-components/src/components/modal/Modal.test.ts
+++ b/web-components/src/components/modal/Modal.test.ts
@@ -44,25 +44,6 @@ describe("Modal Component", () => {
   });
   afterEach(fixtureCleanup);
 
-  test("should set first focusable element if modal open", async () => {
-    jest.useFakeTimers();
-
-    element.show = true;
-    await elementUpdated(element);
-
-    element.backDrop.dispatchEvent(new Event("transitionend"));
-    await nextFrame();
-    await nextFrame();
-    await elementUpdated(element);
-
-    const input = element.querySelector("input");
-
-    jest.runAllTimers();
-
-    expect(document.activeElement).toEqual(input);
-    jest.clearAllTimers();
-  });
-
   test("should dispatch event when modal closed", async () => {
     jest.useRealTimers();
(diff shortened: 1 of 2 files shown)


web-components/src/components/modal/Modal.ts
```diff
@@ -163,7 +163,7 @@ export class Modal extends FocusTrapMixin(LitElement) {
     this.deactivateFocusTrap!();
 
     setTimeout(() => {
-      this.closeModal();
```
futurethang, line 166: This appears to be the only location where `private closeModal()` is used, so the function can be deleted from the file as well from line 112

futurethang (changes requested):

web-components/src/components/modal/Modal.ts
helen-v: Agree! Deleted.

web-components/src/components/modal/Modal.test.ts
```diff
@@ -216,6 +214,5 @@ describe("Modal Component", () => {
     jest.runAllTimers();
 
     expect(document.activeElement).not.toEqual(input);
```
futurethang, line 216: These tests might need rethinking. `document.activeElement()` cannot penetrate the shadowDom, and will always return the outermost custom element, and so this test assertion will always return true.  You'd either need to test in a different way, or replace the native `activeElement` with something that can recursively search into shadowDoms.

futurethang (changes requested):

web-components/src/components/modal/Modal.test.ts
helen-v: Kevin, I agree. And when Kostay will available I'll ask him about logic, Or rewrite tests for modal component.

yananym (approved):

yananym (approved):

PR #704, opened 2020-11-25
fix(Editable-Textfield): Missing max-lines feature: CAX-830

futurethang:
## Description
This fix replaces a missing style calculation that allows user constraint of number of lines.

## Related Issue

https://jira-eng-sjc12.cisco.com/jira/browse/CAX-830

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

- [X] All new and existing tests passed.


helen-v (changes requested):
As I understand it's not work. You don't need max-height, because height can be more as you need. 
editable

As for me, in this case is better use : `display: -webkit-box; -webkit-line-clamp: ${this.maxLines}; word-break: break-all; -webkit-box-orient: vertical; overflow: hidden;`
editable2

futurethang:
> display: -webkit-box; -webkit-line-clamp: ${this.maxLines}; word-break: break-all; -webkit-box-orient: vertical; overflow: hidden;
@helen-v 
That looks better initially, but you can no longer scroll to see the extended text, and the ellipses remain on the second line if you arrow-navigate the typing  cursor down, so not 100% perfect. 

That said, this is very much an emergency replacement of previous functionality, which is what you see currently. It might have room for improvement, but as they currently have it the text box just keeps growing and growing and breaking the UI, so we just want to get it back to acceptable for now.

elenagiere:
I believe this is a significant improvement and doesn't break the entire UI. I noticed that the ellipsis and tooltip is an expectation for bug completion and should be completed, but can be done in a separate PR.

PR #751, opened 2021-01-14
fix(DateTimePicker): fix month navigation, initial values, input field modification

elenagiere:
Addressing a number of bugs found with DateTimePicker
- next/previous month button doesn't change the month header
- added calendar icon on the left of input field
- fixed init value for dateValue to appear properly in input field and in the dateTimePicker
- Made sure that the dateTime can be modified from the value prop and from the input field
- fixed the orange font and orange background when user focuses on today's date.
- Validate the value within the Input field. If not valid, input field is red.

## Description
Bugs:
- next/previous month button doesn't change the month header
- added calendar icon on the left of input field
- fixed init value for dateValue to appear properly in input field and in the dateTimePicker
- Made sure that the dateTime can be modified from the value prop and from the input field

https://jira-eng-chn-sjc3.cisco.com/jira/browse/CAX-878

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


futurethang (commented):
@elenagiere couple things I noticed in usage . . . .
1. When typing  in a date into date-picker with a date range limitation (solo version without time), I am able to enter a date that lies outside of the allowed dates without error.
2. In combined date/time  picker, if the already  selected date is selected again (second click or enter, e.g.) then it causes red error highlighting.

@futurethang Thanks! I have addressed these items.

web-components/src/components/input/Input.ts
```diff
@@ -469,7 +470,7 @@ export class Input extends FocusMixin(LitElement) {
   }
 
   messagesTemplate() {
-    return this.messages && !!this.messages.length
+    return !this.hideMessage && this.messages && !!this.messages.length
```
yananym, line 473: @futurethang did we have this achieved differently?

web-components/src/components/datepicker/DatePicker.ts
```diff
@@ -129,19 +167,35 @@ export class DatePicker extends LitElement {
   chosenDateLabel = () => {
     return this.selectedDate
       ? `, Selected date is ${this.selectedDate.weekdayLong} ${this.selectedDate.monthLong} ${this.selectedDate.day}, ${this.selectedDate.year}`
-      : null;
+      : undefined;
   };
 
-  render() {
-    const placeholder = this.placeholder || this.selectedDate.toLocaleString({ locale: this.locale });
+  isValueValid = () => {
```
yananym, line 173: just in general anything tht starts with an "is" should return a true/false :)

web-components/src/components/input/Input.ts
elenagiere: @yananym with TimePicker, I was able to just use css `::parts` to `display: none` on `md-input::part(message). But in this case, it doesn't seem possible as it is double nested. I tried md-menu-overlay md-input::part(message) but seems like parts doesn't work through two shadowDoms

futurethang (approved):

web-components/src/components/input/Input.ts
yananym: yeah you do it th same way Kevin did in Editable field. makes sense

futurethang (approved):

PR #759, opened 2021-01-26
feat(form): create form component

boiarchukk:
Create new Form component
## Description

Create new Form component

## Types of changes

- [x] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


web-components/src/components/form/Form.ts
```diff
@@ -0,0 +1,189 @@
...
+
+      form.addEventListener("submit", this.handleSubmit);
+
+      const { parentNode: parent } = element;
+
+      if (!element.id) {
+        element.id = "input-form";
+      }
+      if (!element.name) {
+        element.name = "input-form";
```
yananym, line 138: this actually needs to be configurable. only when i change this value to "name" in the DOM manually, I get the autofill window: 
Screen Shot 2021-01-26 at 12 53 31 PM

web-components/src/components/form/Form.ts
```diff
@@ -0,0 +1,189 @@
...
+  @customElement("md-form")
+  export class ELEMENT extends SlottedMixin(LitElement) {
+    @property({ type: String }) rel = "";
+    @property({ type: String }) name = "";
+    @property({ type: String }) action = "";
+    @property({ type: String }) label = "";
+    @property({ type: String }) method: Method = "GET";
+    @property({ type: String }) target: Target = "_self";
+    @property({ type: String }) enctype: Enctype = "application/x-www-form-urlencoded";
+    @property({ type: Boolean, reflect: true }) novalidate = false;
```
elenagiere, line 27: I believe when we have a boolean property that is more than 1 word, we make sure to add an attribute value that provides a camelcase variable name. ex. `no-validate`.

web-components/src/[sandbox]/examples/form.ts
```diff
@@ -0,0 +1,18 @@
+import "@/components/form/Form";
+import { html } from "lit-element";
+
+export const formTemplate = html`
+  <h3>Default</h3>
+  <md-form is-valid>
+    <md-input></md-input>
+    <md-button type="submit">Submit</md-button>
```
Comment (elenagiere) [inline web-components/src/[sandbox]/examples/form.ts:8]: Why have this button here? It doesn't do anything in sandbox.

Also, for this default form, should the entire page refresh when the user invokes the enter key on the input field?
Also, maybe there should be a custom event that fires? or is it the Input web component that will handle that?

Another thing, we should include an example in sandbox for a form with label/input pairing and confirm that the result is in a format similar to `name=Hi&lname=Do `.

elenagiere (commented):
Left comments above.

web-components/src/[sandbox]/examples/form.ts
Comment (boiarchukk) [reply web-components/src/[sandbox]/examples/form.ts:8]: @elenagiere thx for good questions. 

> should the entire page refresh when the user invokes the enter key on the input field?

This is native behaviour of any form that contain [submittable](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#validating_forms_using_javascript) element. 

> Also, maybe there should be a custom event that fires?

For this purpose, my point that native event bubbling will be enough.

> Another thing, we should include an example in sandbox for a form with label/input pairing and confirm that the result is in a format similar to name=Hi&lname=Do 

This will be useful - will do

web-components/src/components/form/Form.ts
boiarchukk: Agreed

boiarchukk: @elenagiere change property name to more appropriate

boiarchukk:
@elenagiere after conversation with @yananym, we resolved that the data is appended to the URL by sending AJAX request or submit form outside of component. I created few prototypes that was a proof of concept implementation, and from my point of view the best solution of "how" and "when" this form will be submitted, is rely on developer requirements.

web-components/src/components/form/Form.ts
yananym: So i added this myself (name attribute to form). Please take notes next time we are discussing action items.

yananym (approved):

PR #787, opened 2021-02-24
feat(spacelistmeeting): add props for button tabIndex and className

prsaroha:
Adding a tabIndex prop and a className prop for the OBTP button.

## Description
We need to make the tabIndex of the OBTP button -1, so adding a prop to allow that while keeping 0 as the default value to preserve old behavior. We also need to pass classes sometimes to the button, so a prop was needed for that as well.

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


dalipkumar703 (approved):

yjlo (commented):
Maybe we should just pass a buttonProps object instead?

prsaroha:
> Maybe we should just pass a buttonProps object instead?

I was thinking that, that being a major change requiring refactor for other users of momentum if they update the version, this may not get merged ASAP. If that's not really a concern, I can make the changes.

yjlo (approved):

robstax (approved):

Issue #827, opened 2021-03-17
Add US flag to phone input country code dropdown box


ccorCis:
In the PhoneInput web component, the US flag picture was added to the country code dropdown box.

## Issue Type
Feature Request

## Component or Pattern Affected

<PhoneInput />

## Expected Behavior

Desired Outcome in image below:
<img width="359" alt="countryFlag" src="https://user-images.githubusercontent.com/79867310/111530037-d6fcbe00-8730-11eb-8334-b0aad64dadfb.png">

## Current Behavior

<img width="385" alt="noFlag" src="https://user-images.githubusercontent.com/79867310/111530735-a8cbae00-8731-11eb-885e-37576086df77.png">

futurethang:
Closed with #855

PR #828, opened 2021-03-17
Add usa flag to PhoneInput country code dropdown combobox

ccorCis:
## Description

In the PhoneInput component there is a combobox that lists country codes. This adds the usa flag when the US country code is selected

## Related Issue
https://github.com/momentum-design/momentum-ui/issues/827

## How Has This Been Tested?

New Unit tests Added. Changes tested in Firefox and Chrome browser.

## Screenshots:
**Before** (If applicable):

noFlag

**After**:

countryFlag

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/web-components/src/components/phone-input/PhoneInput.ts
+++ b/web-components/src/components/phone-input/PhoneInput.ts
@@ -28,7 +28,7 @@ export namespace PhoneInput {
 
   @customElementWithCheck("md-phone-input")
   export class ELEMENT extends LitElement {
-    @property({ type: String }) codePlaceholder = "+1";
+    @property({ type: String, reflect: true }) codePlaceholder = "+1";
     @property({ type: String }) flagClass = "flag__usa";
     @property({ type: String }) numberPlaceholder = "Enter Phone Number";
     @property({ type: String, attribute: "country-calling-code" }) countryCallingCode = "";
@@ -51,6 +51,22 @@ export namespace PhoneInput {
       });
     }
 
+    attributeChangedCallback(name: string, oldval: any, newval: any) {
+      super.attributeChangedCallback(name, oldval, newval);
+      if(name.toLowerCase() === 'codePlaceholder'.toLowerCase()) {
+        this.flagClass = newval === '+1' ? 'flag_usa' : '';
+      }
+    }
+
+    updated(changedProperties:any) {
+      changedProperties.forEach((oldValue:any, propName:any) => {
+        console.log({propName, x: this});
+        if(propName === 'countryCallingCode'){
+          this.flagClass = this.countryCode === 'US' ? 'flag_usa' : '';
+        }
+      });
+    }
+
     static get styles() {
       return [reset, styles];
     }
@@ -70,17 +86,13 @@ export namespace PhoneInput {
     }
 
     handleCountryChange(event: CustomEvent) {
-      if (!event.detail.value) {
-        this.flagClass = 'flag__usa';
+      this.flagClass = this.codePlaceholder === '+1' ? 'flag__usa' : '';
+      if (!event.detail.value || !event.detail.value.id) {
         return;
       }
       this.countryCallingCode = event.detail.value.id;
       this.countryCode = event.detail.value.id.split(",")[2];
-      if(this.countryCode === "US"){
-        this.flagClass = 'flag__usa';
-      } else {
-        this.flagClass = '';
-      }
+      this.flagClass = this.countryCode === "US" ? 'flag__usa' : '';
     }
 
     handlePhoneChange(event: CustomEvent) {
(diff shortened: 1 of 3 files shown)


web-components/src/components/phone-input/PhoneInput.ts
```diff
@@ -23,11 +23,13 @@ export namespace PhoneInput {
...
     value: string;
     errorMessage: string;
+    flagClass: string;
   };
 
   @customElementWithCheck("md-phone-input")
   export class ELEMENT extends LitElement {
-    @property({ type: String }) codePlaceholder = "+1";
+    @property({ type: String, reflect: true }) codePlaceholder = "+1";
+    @property({ type: String }) flagClass = "flag__usa";
```
futurethang, line 32: consider this an element attribute, so be sure that the expected value input is know / intuitive to a user for a range of countries they may set this to, otherwise we should conceal it as an internal property.

web-components/src/components/phone-input/PhoneInput.ts
```diff
@@ -49,6 +51,21 @@ export namespace PhoneInput {
       });
     }
 
+    attributeChangedCallback(name: string, oldval: any, newval: any) {
+      super.attributeChangedCallback(name, oldval, newval);
+      if(name.toLowerCase() === 'codePlaceholder'.toLowerCase()) {
+        this.flagClass = newval === '+1' ? 'flag__usa' : '';
```
futurethang, line 57: We want to move away from any explicit country selection, have it respond to user settings for their own Phone Input usage

web-components/src/components/phone-input/PhoneInput.ts
```diff
@@ -49,6 +51,21 @@ export namespace PhoneInput {
...
+      }
+    }
+
+    updated(changedProperties:any) {
+      changedProperties.forEach((oldValue:any, propName:any) => {
+        if(propName === 'countryCallingCode'){
+          this.flagClass = this.countryCode === 'US' && (this.countryCallingCode.startsWith('+1') || (this.countryCallingCode === '' && this.codePlaceholder == '+1')) ? 'flag__usa' : '';
```
futurethang, line 64: Same scenario about being too specific here. Needs to fit all user scenarios.

web-components/src/components/phone-input/PhoneInput.ts
```diff
@@ -68,11 +85,13 @@ export namespace PhoneInput {
     }
 
     handleCountryChange(event: CustomEvent) {
-      if (!event.detail.value) {
+      this.flagClass = this.codePlaceholder === '+1' ? 'flag__usa' : '';
+      if (!event.detail.value || !event.detail.value.id) {
```
futurethang, line 89: Same idea on specificity vs universality

web-components/src/components/phone-input/PhoneInput.ts
```diff
@@ -68,11 +85,13 @@ export namespace PhoneInput {
...
         return;
       }
       this.countryCallingCode = event.detail.value.id;
       this.countryCode = event.detail.value.id.split(",")[2];
+      this.flagClass = this.countryCode === "US" ? 'flag__usa' : '';
```
futurethang, line 94: Same idea on specificity vs universality

web-components/src/components/phone-input/PhoneInput.ts
```diff
@@ -127,16 +146,23 @@ export namespace PhoneInput {
       this.formattedValue = new AsYouType(this.countryCode).input(input);
     }
 
+    handleCountryFocusOut({detail}:any){
+      this.flagClass = (detail.value === '' && this.codePlaceholder === '+1') || (detail.value === '+1' && this.countryCode === 'US' && this.countryCallingCode.startsWith('+1')) ? 'flag__usa' : ''
```
futurethang, line 150: Same idea on specificity vs universality

web-components/src/components/phone-input/scss/phoneInput.scss
```diff
@@ -3,6 +3,21 @@
   display: flex;
   position: relative;
 
+  md-combobox.flag__usa {
+    &::part(multiwrap) {
+      &:before {
```
futurethang, line 8: In a more robust, universal solution, this might be approached differently

web-components/src/components/phone-input/PhoneInput.ts
```diff
@@ -49,6 +51,21 @@ export namespace PhoneInput {
       });
```
futurethang, line 51: In order for this to not be a breaking change, we need another attribute setting for users to explicitly opt-in to a `flags` feature so that all current users who like it the way it is do not have to update their software.
it could be like 
`@property({ type: Boolean, attribute: "show-flags" }) showFlags = false;`

futurethang (changes requested):
It's a start in the right direction! In order to adopt the changes to mainstream Momentum WC we will have to ensure users can choose whether or not to use the extra feature (assume opt-out) and then flexibly cover all country flag scenarios.

futurethang:
https://github.com/madebybowtie/FlagKit

https://www.npmjs.com/package/country-code-emoji

PR #837, opened 2021-03-30
fix(floating-modal): minWidth & minHeight CAX-931

Budhia:
## Description

Added minWidth and minHeight as properties to the md-floatingModal

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/web-components/src/components/floating-modal/FloatingModal.stories.ts
+++ b/web-components/src/components/floating-modal/FloatingModal.stories.ts
@@ -48,11 +48,13 @@ export const FloatingModal = () => {
   const show = boolean("show", false);
   const full = boolean("full-screen", false);
   const fixed = boolean("fixed-strategy", false)
+  const minWidth = text("min-Width", "");
+  const minHeight = text("min-Height", "");
 
   return html`
 
     <md-theme class="theme-toggle" id="floating" ?darkTheme=${darkTheme} ?lumos=${lumos}>
-    <md-floating-modal ?show=${show} ?full-screen=${full} ?fixed-strategy=${fixed}>
+    <md-floating-modal ?show=${show} ?full-screen=${full} ?fixed-strategy=${fixed} .minWidth=${minWidth} .minHeight=${minHeight}>
       <md-radiogroup group-label="group_process">
         <md-radio slot="radio" value="Option 1">Option 1</md-radio>
         <md-radio slot="radio" value="Option 2">Option 2</md-radio>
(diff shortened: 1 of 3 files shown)


web-components/src/components/floating-modal/FloatingModal.stories.ts
```diff
@@ -48,11 +48,13 @@ export const FloatingModal = () => {
...
   const full = boolean("full-screen", false);
   const fixed = boolean("fixed-strategy", false)
+  const minWidth = text("min-Width", "");
+  const minHeight = text("min-Height", "");
 
   return html`
 
     <md-theme class="theme-toggle" id="floating" ?darkTheme=${darkTheme} ?lumos=${lumos}>
-    <md-floating-modal ?show=${show} ?full-screen=${full} ?fixed-strategy=${fixed}>
+    <md-floating-modal ?show=${show} ?full-screen=${full} ?fixed-strategy=${fixed} .minWidth=${minWidth} .minHeight=${minHeight}>
```
futurethang, line 57: Following the comment below about kebab case, let's also make these `min-width` and `min-height`

web-components/src/components/floating-modal/FloatingModal.ts
```diff
@@ -23,6 +23,8 @@ export namespace FloatingModal {
     @property({ type: Boolean, reflect: true, attribute: "aspect-ratio" }) aspectRatio = false;
     @property({ type: Boolean, reflect: true, attribute: "fixed-strategy" }) fixed = false;
     @property({ type: Boolean, reflect: true, attribute: "full-screen" }) full = false;
+    @property({ type: String }) minWidth = "";
+    @property({ type: String }) minHeight = "";
```
futurethang, line 27: Let's keep up with exposing kebab-case attribute names, like so:
```
 @property({ type: String, attribute: "min-width" }) minWidth = "";
 @property({ type: String, attribute: "min-height" }) minHeight = "";
    ```

futurethang (commented):
Read my comments because they apply to many cases of adding sttributes, but for this particular use case, I don't believe we need to make changes to the component, we just need to make the docs more clear that you can in fact achieve this effect with CSS. To demonstrate (and see if it matches your desired outcome) put the following in the floatin-modal sandbox file, just within the html template on line 33:
```js
<style>
        md-floating-modal::part(floating) {
          min-height: 400px;
          min-width: 600px;
        }
      </style>
      ```

PR #845, opened 2021-04-07
fix(web-components): accessibility

Budhia:
## Description

- Added a property ariaLabel to md-combobox, md-editable-field, md-modal
- Removed Menu-overlay close trigger on pressing Enter. The Enter button will now function to perform internal functions inside the menu-overlay. MenuOverlay can be closed using Esc or directly from the component side using reference.
- md-menu-overlay opens on Enter, ealier it used to open only on Space
- Agent history tabpanel - Rectified the below described issue
  The focus should be always on the current selected tab. Current logic brings focus on the last navigated tab.	-i.e When the current tab is “All” (same for other tabs), press Tab to navigate to the tablist then press left arrow to move the “Chat” tab (without activating it), then continue to press Tab to navigate. When the focus moves back to the tablist, the focus is on the “Chat” tab, not the current tab “All”
- alertbanner component : Updated closebutton ariaLabel to 'Close banner to be more meaningfull'>
- Tooltip Component : Minimum width of the tootltip was removed to adjust to the perfect text without any extra space. 

## Motivation and Context
https://jira-eng-sjc12.cisco.com/jira/browse/CAX-932

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/web-components/src/components/alert-banner/AlertBanner.ts
+++ b/web-components/src/components/alert-banner/AlertBanner.ts
@@ -56,7 +56,7 @@ export namespace AlertBanner {
             <md-button
               class="md-alert-banner__close"
               hasRemoveStyle
-              ariaLabel="Close"
+              ariaLabel="Close Banner"
               @click="${this.onHide}"
               @keydown="${this.handleKeyDown}"
             >
--- a/web-components/src/components/alert-banner/AlertBanner.ts
+++ b/web-components/src/components/alert-banner/AlertBanner.ts
@@ -22,6 +22,7 @@ export namespace AlertBanner {
     @property({ type: String }) message = "";
     @property({ type: Boolean }) closable = false;
     @property({ type: Boolean }) show = false;
+    @property({ type: String, attribute: "close-aria-label" }) closeAriaLabel = "Close Banner";
 
     connectedCallback() {
       super.connectedCallback();
@@ -56,7 +57,7 @@ export namespace AlertBanner {
             <md-button
               class="md-alert-banner__close"
               hasRemoveStyle
-              ariaLabel="Close Banner"
+              ariaLabel=${this.closeAriaLabel}
               @click="${this.onHide}"
               @keydown="${this.handleKeyDown}"
             >
--- a/web-components/src/components/combobox/ComboBox.ts
+++ b/web-components/src/components/combobox/ComboBox.ts
@@ -59,6 +59,7 @@ export namespace ComboBox {
     @property({ type: Boolean, reflect: true, attribute: "search-trim-space" }) trimSpace = false;
     @property({ type: Boolean, reflect: true }) invalid = false;
     @property({ type: String, reflect: true, attribute: "invalid-text-i18n" }) invalidText = "";
+    @property({ type: String }) ariaLabel = "";
 
     @property({ type: Number, attribute: false })
     get focusedIndex() {
@@ -774,7 +775,7 @@ export namespace ComboBox {
         <button
           type="button"
           class="md-combobox-button clear"
-          aria-label=${this.label}
+          aria-label="Clear button"
           aria-expanded=${this.expanded}
           aria-controls="md-combobox-listbox"
           tabindex="-1"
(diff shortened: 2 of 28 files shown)


web-components/src/components/combobox/ComboBox.ts
```diff
@@ -59,6 +59,7 @@ export namespace ComboBox {
     @property({ type: Boolean, reflect: true, attribute: "search-trim-space" }) trimSpace = false;
     @property({ type: Boolean, reflect: true }) invalid = false;
     @property({ type: String, reflect: true, attribute: "invalid-text-i18n" }) invalidText = "";
+    @property({ type: String }) ariaLabel = "";
```
elenagiere, line 62: When there are properties with camelCase, we want to add an attribute value with it in dashes like so
`@property({ type: String, attribute: "aria-label" }) ariaLabel = "";`

elenagiere, line 62: When there are properties with camelCase, we want to add an attribute value with it in dashes like so
`@property({ type: String, attribute: "aria-label" }) ariaLabel = "";`
This also changes how you pass it in.

web-components/src/components/menu-overlay/MenuOverlay.ts
```diff
@@ -287,7 +287,7 @@ export namespace MenuOverlay {
         return;
       }
 
-      if (event.code === Key.Escape || event.code === Key.Enter) {
+      if (event.code === Key.Escape) {
```
elenagiere, line 290: I believe there are cases in which people want it to close when entered. So maybe we can add a boolean property in which when true, it will close on enter? Imagine a list and when you select an item in the list, the menuOverlay closes. Just thought we could add the flexibility to add that functionality back in if wanted.

futurethang: I second this

Budhia: Sure will incorporate the changes

Budhia: I looked into it. So currently what is happening is if you press Enter on the item in the list (according to the example you gave above) nothing will happen. It just closes. It will not select and then close. The bug was raised because of this only. Because on the desktop side people wanted the menu overlay to close upon pressing Enter on Submit but right now it is just closing and not performing the action on Submit. 
At every place in the desktop after the Enter they are closing the modal using ref. 
So according to me if we are giving Enter using the boolean property it will be useful only in the situation where user has opened menu overlay just to read the content and not perform any action.

elenagiere: Ok

web-components/src/components/alert-banner/AlertBanner.ts
```diff
@@ -56,7 +56,7 @@ export namespace AlertBanner {
             <md-button
               class="md-alert-banner__close"
               hasRemoveStyle
-              ariaLabel="Close"
+              ariaLabel="Close Banner"
```
merajcal, line 59: why hard coded ariaLabel? it should be passed from application. so that it can be localized

web-components/src/components/combobox/ComboBox.ts
```diff
@@ -774,15 +775,15 @@ export namespace ComboBox {
         <button
           type="button"
           class="md-combobox-button clear"
-          aria-label=${this.label}
+          aria-label="Clear button"
```
merajcal, line 778: why hard coded ariaLabel? it should be passed from application. so that it can be localized

web-components/src/components/combobox/ComboBox.ts
```diff
@@ -793,7 +794,7 @@ export namespace ComboBox {
         <button
           type="button"
           class="md-combobox-button arrow-down"
-          aria-label=${this.label}
+          aria-label="Arrow Down"
```
merajcal, line 797: why hard coded ariaLabel? it should be passed from application. so that it can be localized

web-components/src/components/combobox/ComboBox.ts
```diff
@@ -819,7 +820,7 @@ export namespace ComboBox {
                 type="text"
                 role="combobox"
                 aria-autocomplete="both"
-                aria-label="input"
+                aria-label="Combobox Input"
```
merajcal, line 823: why hard coded ariaLabel? it should be passed from application. so that it can be localized

merajcal (changes requested):

elenagiere (dismissed):

merajcal (approved):

futurethang (approved):

Issue #848, opened 2021-04-08
Tooltip text is vertical in lightbox react component


Coread:
## Issue Type
Bug Report

## Component or Pattern Affected
Lightbox - react

## Expected Behavior
Text in tooltips should be written horizontally

## Current Behavior
Text is vertical

## Steps to Reproduce (for bugs)
See kitchen sink lightbox component, exit and next tooltips

![lightbox_before](https://user-images.githubusercontent.com/3998548/114013686-b9b69d80-985f-11eb-85b5-9631bf440c91.png)

PR #849, opened 2021-04-08
fix(tooltips): fix tooltips for lightbox

Coread:
Fix a few issues with the event overlay which are noticeable on the lightbox tooltips

## Description
When the isContained flag is true, both bottom and left positioned overlays were broken. The bottom positioning was broken when the element is on the far right of the page. The left positioned overlay would stretch to the far left of the page.

In addition, the props for the tooltips in the lightbox were being specified incorrectly. The Tooltip component does not take a direction property, instead taking a popoverProps object.

In the EventOverlay component, reliance on documentRight for positioning is broken in the presence of a scrollbar. I've removed this problem by no longer relying on documentRight for the bottom or left contained positioning.

## Related Issue
#848

## Motivation and Context
Fixes light box tooltips

## How Has This Been Tested?
Checked visually on the lightbox and tooltip kitchen sink components along with every other component that uses lightbox. I've also added some more cases to the tooltip direction kitchen sink

## Screenshots:
**Before**:
lightbox_before

**After**:
lightbox_after

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/react/src/lib/EventOverlay/index.js
+++ b/react/src/lib/EventOverlay/index.js
@@ -472,7 +472,7 @@ class EventOverlay extends React.Component {
     const documentScrollTop = document.documentElement.scrollTop;
     const documentBottom = document.documentElement.scrollHeight;
     const windowBottom = window.pageXOffset + window.innerHeight;
-    const documentRight = document.documentElement.offsetWidth;
+    const documentRight = Math.max(document.documentElement.offsetWidth, document.documentElement.clientWidth);
     const arrowHeight = (arrowDims && arrowDims.height) || 0;
     const arrowWidth = (arrowDims && arrowDims.width) || 0;
     const offsetHeight = targetOffset.vertical || 0;
@@ -610,7 +610,7 @@ class EventOverlay extends React.Component {
           if (elementDims.right >= documentRight || this.elementWidth > documentRight) {
             targetNode.style.right = '0px';
             if (this.elementWidth < documentRight) {
-              targetNode.style.left = 'revert';
+              targetNode.style.left = 'inherit';
             }
           }
           if (elementDims.left < 0) {
@@ -686,19 +686,14 @@ class EventOverlay extends React.Component {
         break;
       case 'left':
         if (!scrollParentDims.left && !transformParentDims) {
-          if (
-            arrowWidth + offsetWidth + elementDims.width + anchorPosition.left >
-            anchorPosition.left
-          ) {
-            if (anchorPosition.left  - offsetWidth - elementDims.width > arrowWidth) {
-              targetNode.style.left = 'revert';
-            } else {
-              targetNode.style.left = `${arrowWidth}px`;
-            }
-            targetNode.style.right = `${documentRight -
-              anchorPosition.left +
-              arrowWidth +
-              offsetWidth}px`;
+          targetNode.style.right = `${documentRight -
+            anchorPosition.left +
+            arrowWidth +
+            offsetWidth}px`;
+          if ( elementDims.left - arrowWidth < 0 ) {
+            targetNode.style.left = `${arrowWidth}px`;
+          } else {
+            targetNode.style.left = 'inherit';
           }
           if (getAvailableTopSpace(documentScrollTop) < 0) {
             targetNode.style.top = `${-documentScrollTop}px`;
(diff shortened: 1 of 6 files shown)


futurethang:
I think this looks good, but want to be sure we consider, are the changes in EventOverlay friendly for all usages, not just ToolTips? I know that that component is utilized in a lot of different ways, so there's a stronger concern for potentially breaking changes. Have you tested other components that might rely upon this mechanic of EventOverlay?

react/src/lib/EventOverlay/index.js
```diff
@@ -472,7 +472,7 @@ class EventOverlay extends React.Component {
     const documentScrollTop = document.documentElement.scrollTop;
     const documentBottom = document.documentElement.scrollHeight;
     const windowBottom = window.pageXOffset + window.innerHeight;
-    const documentRight = document.documentElement.offsetWidth;
+    const documentRight = Math.max(document.documentElement.offsetWidth, document.documentElement.clientWidth);
```
Coread, line 475: The document width is altered by the no-scroll plugin which makes documentRight wrong by a scrollbar width. Using the max of offset and client width stops this happening

react/src/lib/EventOverlay/index.js
```diff
@@ -607,10 +607,10 @@ class EventOverlay extends React.Component {
           ) {
             targetNode.style.bottom = `${documentScrollTop + windowBottom - documentBottom}px`;
           }
-          if (elementDims.right > documentRight || this.elementWidth > documentRight) {
+          if (elementDims.right >= documentRight || this.elementWidth > documentRight) {
```
Coread, line 610: By this point in the code the overlay box has already been placed. If it is right at the edge of the page it will have the same right value as documentRight hence the change to >=

react/src/lib/EventOverlay/index.js
```diff
@@ -607,10 +607,10 @@ class EventOverlay extends React.Component {
...
             targetNode.style.right = '0px';
             if (this.elementWidth < documentRight) {
-              targetNode.style.left = `${documentRight - this.elementWidth}px`;
+              targetNode.style.left = 'inherit';
```
Coread, line 613: The elementWidth is unreliable at this point. It has already been placed and if it is constrained by the right hand edge the box is likely too narrow (e.g. the vertical exit tooltip). Here I unset the left css rule so the tooltip can be placed naturally

react/src/lib/EventOverlay/index.js
```diff
@@ -686,15 +686,14 @@ class EventOverlay extends React.Component {
...
       case 'left':
         if (!scrollParentDims.left && !transformParentDims) {
-          if (
-            arrowWidth + offsetWidth + elementDims.width + anchorPosition.left >
-            anchorPosition.left
-          ) {
+          targetNode.style.right = `${documentRight -
+            anchorPosition.left +
+            arrowWidth +
+            offsetWidth}px`;
```
Coread, line 692: We always need to specify the right hand position here. the previous if check was pretty much always true so for the majority of cases the logic hasn't changed.

react/src/lib/EventOverlay/index.js
```diff
@@ -686,15 +686,14 @@ class EventOverlay extends React.Component {
...
+          if ( elementDims.left - arrowWidth < 0 ) {
```
Coread, line 693: The minimum left placement possible is arrowWidth. This is due to a negative arrowWidth margin applied in the common css for left hand placed overlays.

react/src/lib/EventOverlay/index.js
```diff
@@ -686,15 +686,14 @@ class EventOverlay extends React.Component {
...
             targetNode.style.left = `${arrowWidth}px`;
-            targetNode.style.right = `${documentRight -
-              anchorPosition.left +
-              arrowWidth +
-              offsetWidth}px`;
+          } else {
+            targetNode.style.left = 'inherit';
```
Coread, line 696: Previously the left edge of the overlay would always be at the left edge of the screen. This else takes care of the case where there is enough room to place the tooltip properly, i.e. it is not too far left

Coread:
> I think this looks good, but want to be sure we consider, are the changes in EventOverlay friendly for all usages, not just ToolTips? I know that that component is utilized in a lot of different ways, so there's a stronger concern for potentially breaking changes. Have you tested other components that might rely upon this mechanic of EventOverlay?

I've been through the other components that use the overlay. The main changes in this PR affect only the constrained (left and bottom) placement. I couldn't see any other component that was using the overlays in that way. I've also done plenty of playing around with the kitchen sink for overlay, popover and tooltip.

robstax (approved):

PR #885, opened 2021-05-10
fix(icon-build): ensure that rebuilds of icons produce the same output

paulwitty:
## Description
I got fed up that rebuilding the icon files produced a massive diff every time, even if the inputs hadn't changed. This was because a timestamp (of now) was used in the file. I've changed this to instead use the timestamp of the last git commit in the icons/svg directory, which seems like a much better timestamp to use.

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [x] Build related changes
- [ ] CI related changes

## Checklist:


paulwitty:
@futurethang afraid I haven't fixed that, because I wasn't aware of the problem. I'll see if I can think of a way to make it consistent...

futurethang:
@paulwitty cool cool. for reference, see what `icons.scss` is all about at `momentum-ui/icons/scss/icons.scss`

lismith2-cisco (commented):
I suggest building icons & committing as part of this PR so everything is in a good state

paulwitty:
@futurethang 
Each icon gets mapped to a glyph in the font file, starting from a fixed point and then assigning glyphs sequentially in alphabetical order of icon names. So if you add an icon with a name beginning with a, all the icons get new glyphs, resulting in much bigger diffs.

We could fix that by:
a) choosing glyphs using a hash of the name, so it's consistent, but we'll get collisions so it's only mostly consistent
b) using the previous mapping as an input to the new mapping, but if you remove and then re-add an icon it'll still get a different glyph

But we have an scss file which provides the mapping of name to glyph, and we always use that to refer to icons anyway, so the only benefit we get is reducing the size of diffs (on generated files), and I don't consider that worth putting in the effort

lismith2-cisco (approved):

futurethang:
@paulwitty @lismith2-cisco OK cool let's merge this, but please keep a sharp eye out for any Icon trouble in the chat spaces.

PR #910, opened 2021-05-20
feat(react/themeselect): select theme for kitchen sink

Coread:
Added a theme select drop down to the react kitchen sink. Will make it easier to check theming across the whole kitchen sink.

Hard coded themes at the moment but could eventually come generated from tokens when that is available.

## Screenshots:

theme select

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


react/src/app/ThemeSelect/themes.js
```diff
@@ -0,0 +1,177 @@
+export const light = {
```
InteractiveTimmy, line 1: These should be generated dynamically from `@momentum-ui/tokens` => `semanticColor.json`. See `@cantina/state/theme/utils.js`

InteractiveTimmy (approved):
One, non-blocking comment. It'd be useful to have these dynamically generated from the `@momentum-ui/tokens` package to have them accurately reflect what `@momentum-ui` package consumers will see.

Note: Theming implementation occurs at the `@momentum-ui/core` level, could these changes be implemented there so that they can be used during their development?

react/src/app/ThemeSelect/themes.js
paulwitty: I've discussed with @Coread that we'll do this after the new token scheme goes in, rather than adapting it to the old one and then the new one later.

PR #924, opened 2021-05-27
fix(spinner): add new fix

torgeadelin:
Spinner Fix2

## Description
Previous fix doesn't fix the issue.

## Related Issue
Spinner issue

## Types of changes
- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


InteractiveTimmy (approved):
In a meeting with @torgeadelin it appeared that these changes fixed the spinner issue.

Note: We should be defining all paired mixin parameters as `--null` when they're intended to be CSS Variables per @torgeadelin's discovery.

torgeadelin:
To understand what the issue is here's a codepen to explain the issue: https://codepen.io/torgeadelin/pen/ExWvzEY
null scss variables translate to "" (empty string) and when the scss is built the css code looks like: color: var(,someColor) whici is wrong.

robstax (approved):

Coread (commented):
See the Percy snapshots. The text for placeholder inputs is now unreadable in some cases

torgeadelin:
Because of the introduction of --null instead of the scss null, in mixins, when both the scss variable and css variable are null, the property using them will result in `property: var(--null,)` which breaks the style in some places.

From now one, all changes related to theming in mixins should be wrapped in 
```scss
@if $scssVariable { 
  property: $scssVariable;
  property: var(--css-var, $scssVariable);
}
```
to prevent the whole property to be applied on the element with the broken `property: var(--null,)`.

Again, when you need to change other scss (not mixins) you shouldn't have to wrap the property with `@if` because the scss variable would most likely have a defined values and won't be null (like in mixins).

torgeadelin:
Also, first one seems like it's moved?? It was flaky on my local machine.
Regarding the Percy issues with the icon, I think it's due to timing issues?

Coread (approved):

PR #939, opened 2021-06-10
fix(accessibility): minor changes to EditableTextField and Menu keydown

prsaroha:
Small changes in keydown handling of EditableTextField and Menu components to assist with accessibility.

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/react/src/lib/EditableTextfield/index.js
+++ b/react/src/lib/EditableTextfield/index.js
@@ -66,10 +66,13 @@ class EditableTextfield extends React.Component {
   }
 
   handleEsc = e => {
+    var value = e.target.value;
+    var handleDoneEditing = this.props.handleDoneEditing;
     this.setState(
       {
         isEditing: false
-      }
+      },
+      () => handleDoneEditing && handleDoneEditing(e, {value})
     );
     e.nativeEvent.stopImmediatePropagation();
   }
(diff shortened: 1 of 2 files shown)


react/src/lib/Menu/index.js
```diff
@@ -251,7 +251,8 @@ class Menu extends React.Component {
         flag = true;
         break;
 
-      case 37: //left
+      case 27:
+      case 37: //escape or left
```
prsaroha, line 255: The handling in this case checks for a submenu and then closes it. Currently pressing left key closes the submenu and escape is expected to do the same.

react/src/lib/EditableTextfield/index.js
```diff
@@ -66,10 +66,13 @@ class EditableTextfield extends React.Component {
...
 
   handleEsc = e => {
+    var value = e.target.value;
+    var handleDoneEditing = this.props.handleDoneEditing;
     this.setState(
       {
         isEditing: false
-      }
+      },
+      () => handleDoneEditing && handleDoneEditing(e, {value})
```
prsaroha, line 75: pressing escape while editing is the same as pressing enter without editing, so just like in the case of "enter" key being pressed, escape should also return to the `handleDoneEditing` method if available. Also helps in putting the focus back on this component after editing is done because currently focus gets lost

PR #959, opened 2021-06-22
fix(md-modal): close event firing on first render

sbudhia:
## Description

```close-modal```` is fired even upon when component renders as the else part of the updated section that was having "show" property was not checking anything else

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/web-components/src/components/modal/Modal.ts
+++ b/web-components/src/components/modal/Modal.ts
@@ -10,8 +10,8 @@ import "@/components/button/Button";
 import "@/components/icon/Icon";
 import { Key } from "@/constants";
 import { FocusTrapMixin } from "@/mixins";
-import reset from "@/wc_scss/reset.scss";
 import { customElementWithCheck } from "@/mixins/CustomElementCheck";
+import reset from "@/wc_scss/reset.scss";
 import { html, internalProperty, LitElement, property, PropertyValues, query } from "lit-element";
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by merajcal — APPROVED** [review]

File: web-components/src/components/modal/Modal.ts [hunk 1]
```diff
@@ -10,8 +10,8 @@ import "@/components/button/Button";
 import "@/components/icon/Icon";
 import { Key } from "@/constants";
 import { FocusTrapMixin } from "@/mixins";
-import reset from "@/wc_scss/reset.scss";
```
futurethang, line 13: Why is this removed? Be aware, the reset.scss contains styles that otherwise get messed up because of shadow-dom barriers, like `display`

futurethang (changes requested):

web-components/src/components/modal/Modal.ts
sbudhia: @futurethang  I think my VSCode must have auto-removed. Just few days back installed a plugin which actually auto updated the import and removes the one which are not being used

sbudhia: I will insert it again if you want

futurethang (approved):

PR #962, opened 2021-06-23
fix(md-advanced-table): not rendering from react based repo

sbudhia:
## Description

React based repo are not able to render advanced table due to prop rendering issue. For react based repo calling web-components where prop is an Object it is expected to be sent using JSON.stringify but as we were not accepting string props it was causing issue in react based repos. So what I did is added string support and object both. If we receive prop as a string we are converting it to an Object and then the rest of the process continues as it is without any interruption.

## Related Issue

https://jira-eng-sjc12.cisco.com/jira/browse/CAX-1086?filter=32565

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


web-components/src/components/table-advanced/TableAdvanced.ts
```diff
@@ -68,20 +71,39 @@ export namespace TableAdvanced {
...
-         this.updateDataInTable();
+      if (changedProperties.has("data")) {
+        this.updateTableData();
+        this.updateDataInTable();
+      }
+    }
+
+    private updateTableData() {
+      if (typeof this.data === "string") {
+        this.tableData = JSON.parse(this.data);
```
merajcal, line 89: If there is an invalid format the JSON.parse throws error, but its not very readbale. add try catch and throw parse error along with readable message

web-components/src/components/table-advanced/TableAdvanced.ts
```diff
@@ -68,20 +71,39 @@ export namespace TableAdvanced {
...
+      } else {
+        this.tableData = { ...this.data };
+      }
+    }
+
+    private updateTableConfig() {
+      if (typeof this.config === "string") {
+        this.tableConfig = JSON.parse(this.config);
```
merajcal, line 97: same as above

web-components/src/components/table-advanced/TableAdvanced.ts
```diff
@@ -157,64 +179,63 @@ export namespace TableAdvanced {
...
-           });
-         });
-       }
- 
+      // TEMPLATES
+      const templates = this.tableConfig.cellTemplates;
+      const templatesKeys = Object.keys(templates || {});
+      if (templates && templatesKeys.length) {
+        this.ROWS.forEach((row, iRow) => {
+          row.forEach((cell, iCol) => {
```
merajcal, line 205: can you refactor? what is happening here its not very readable

merajcal (changes requested):

web-components/src/components/table-advanced/TableAdvanced.ts
sbudhia: Will incorporate the try and catch

web-components/src/components/table-advanced/TableAdvanced.ts
sbudhia: Not made any change here. Although here the templates are being populated

web-components/src/components/table-advanced/TableAdvanced.ts
```diff
@@ -68,20 +71,47 @@ export namespace TableAdvanced {
...
+      } else {
+        this.tableConfig = { ...this.config };
       }
-    } 
+    }
 
     private populateColumns() {
       let index = 0;
       const pushCol = (col: ColOptions, group?: { name: string; length: number }) => {
-        const f = col.filters || this.config.default?.col?.filters;
```
futurethang, line 84: is this `config` property still in use elsewhere? It seems like it is still initialized as a prop in the component (like, not removed or replaced explicitly by tableConfig)

sbudhia: It has been replaced everywhere in the logic part of the code so what I am doing is user will still send config and data to props named config and data so as to ensure nothing breaks wherever it is currently running. After that I am taking the value from config and checking if it is string or not and assigning it to tableConfig and similarly to tableData. Then I am using tableConfig and tableData everywhere

futurethang: Ah, so kinda to prevent a breaking change it sounds like. Cool

futurethang (approved):

PR #1025, opened 2021-08-08
CX-6816 : Modal window gets closes on hover of cross button

msnowshadali:
## Description

Modal gets closed when you hover mouse over the cross button 

hover

RCA:
Due to CSS transition on hover of md-close button the modal window is getting closed

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


futurethang:
Is there a bug ticket that this links to and resolves?
I am unable to replicate this bug in any of the Storybook or sandbox environments.

Is there more context, like particular OS or Browser?

My concern overall with this change is that we are rolling back a UI feature that is intentional. So we ought to be sure that it is a wider ranging bug, and then be sure we have exhausted all other attempts at correcting the buggy behavior without removing features. At the moment I am hesitant to delete this UI feature.

sbudhia:
@futurethang 
Bug Ticket : https://jira-eng-sjc12.cisco.com/jira/browse/CX-6816
This bug is getting replicated everywhere in the sandbox and in the storybook. What you need to do is have a modal and keep the cross button on top-right visible. Just take your mouse and hover around it 1-2 times and you will notice the modal get's automatically closed.

The part that has been removed regarding the easy-transition is on the close button and I don't think that will affect the overall UI feature or experience because I also tested it and there was so change at all.

futurethang (approved):

futurethang:
@sbudhia OK cool, thanks for the context there, I can see this more clearly now!

futurethang (approved):

PR #1477, opened 2023-09-12
fix: 🐛 accessibility fixes & focus trap issue of overlay

shrinath-shenoy:
## Description
Addressed few accessibility related issues, fixed keyboard tab focus not working on overlays shown over another overlay and other enhancements

## Motivation and Context
Many of these component do not have option to set accessibility related attributes to the inner components they render.

## How Has This Been Tested?
Major chunk of these changes are basically related to Screen reader accessibility tested with MacOS screen reader. Other changes are manually tested with individual scenarios.

## Types of changes

- [x] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


web-components/src/components/tooltip/Tooltip.ts
```diff
@@ -56,6 +56,16 @@ export namespace Tooltip {
...
     private slotContent: Element[] | null = null;
 
+    connectedCallback() {
+      super.connectedCallback();
+      document.addEventListener("keydown", this.handleKeyDown.bind(this));
+    }
+
+    disconnectedCallback(): void {
+      super.disconnectedCallback();
+      document.removeEventListener("keydown", this.handleKeyDown.bind(this));
```
parsoman, line 66: We should not use bind. Please use arrow function.

web-components/src/mixins/FocusTrapMixin.ts
```diff
@@ -361,33 +375,51 @@ export const FocusTrapMixin = <T extends AnyConstructor<FocusClass & FocusTrapCl
...
+        this.focusableElements = [];
       }
       this.focusableTimer = setTimeout(() => {
         this.setFocusableElements();
       }, 10);
-    }
+    };
+
+    handleChildFocusTrap = (event: CustomEvent) => {
+      if (event.target !== this) {
```
parsoman, line 390: Why this is required. Please explain the use case.

shrinath-shenoy: The use case is, when we try to show a overlay on top of another where the focus is trapped ( e.g. Calendar control over a Modal ) the focus is always trapped on to the modal and not the uppermost calendar. With these changes, as soon as the calendar acquires focus trap, the modal will forgoes the focus trap based on the event emitted by child - calendar. As soon as child forgoes focus, the parent if any will re-acquire the focus.

Since this method is on generic mixin, now when the calendar emits focus-untrap event the same should not be consumed by itself as it also listens to the same event and assume some other child has forgone the focus and re-acquire itself. Hence this check to avoid such cases.

prakhar-tiwari: Have we tested other use cases as well? If modal contains some buttons along with Menu Overlay, once menu overlay is opened, will the focus be trapped inside Menu Overlay only? As there can be a use case where we may need to trap focus only inside Modal

shrinath-shenoy: Yes. These scenarios have been tested

prakhar-tiwari (dismissed):

parsoman (approved):

web-components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@momentum-ui/web-components",
-  "version": "2.7.47",
+  "version": "2.7.49",
```
parsoman, line 3: It should be 2.7.48 , right?

shrinath-shenoy: I see another PR here https://github.com/momentum-design/momentum-ui/pull/1482/files. So made it 2.7.49 to avoid conflicts..

parsoman: have you not taken the latest master in your branch. Please take latest master.

parsoman: You dont have to worry about other PR, if your PR needs to be merged and published it would be 1 more than previous. Other PR owners will update their PR accordingly.

shrinath-shenoy: Done.. Updated the version

prakhar-tiwari (approved):

PR #1561, opened 2024-03-11
#CX-10274: Passing List Items Count To The Parent Fix

mithun-srinivas:
## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1604, opened 2024-04-29
fix(cx-14211): use property name as ariaLabel

neelabhnagaich:
currently we are passing aria-label with the same semantic name as 'aria-label' due to which screen reader is announcing at md-combobox level which is not expected hence changing it to not use the same pre defined name for property.

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


vinitkumargoel (dismissed):

web-components/src/components/combobox/ComboBox.ts
```diff
@@ -1557,7 +1558,6 @@ export namespace ComboBox {
       return html`
         <div
           part="combobox"
-          aria-label=${this.ariaLabelForComboBox}
```
niti-owner, line 1560: Why this is removed? 
Just verify if there is no impact on below PR (CX-10274: Passing List Items Count To The Parent Fix)?
https://github.com/momentum-design/momentum-ui/pull/1561

web-components/src/[sandbox]/examples/combobox.ts
```diff
@@ -25,7 +25,7 @@ export const comboBoxTemplate = html`
     style="width: 220px; display: inline-block;"
     .options=${comboBoxOptions}
     placeholder="Placeholder"
-    aria-label="Select the country"
+    ariaLabel="Select the country"
     search-result-aria-label="Select the country, {{count}} results found."
   ></md-combobox>
   <h3>Default with trim space</h3>
```
Comment (niti-owner) [inline web-components/src/[sandbox]/examples/combobox.ts:31]: Can you also add one example with comboboxId ?

web-components/src/components/combobox/ComboBox.ts
neelabhnagaich: this was extra label on the parent container, which unnecessarily added, hence removed

web-components/src/[sandbox]/examples/combobox.ts
Comment (neelabhnagaich) [reply web-components/src/[sandbox]/examples/combobox.ts:44]: done

merajcal (approved):

niti-owner (approved):

merajcal (approved):

PR #1626, opened 2024-05-22
#CX-13430 :  modal title screen reader fix

binduN:
## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/web-components/src/components/modal/Modal.ts
+++ b/web-components/src/components/modal/Modal.ts
@@ -39,7 +39,6 @@ export namespace Modal {
     @property({ type: String }) headerMessage = "";
     @property({ type: String }) htmlId = "";
     @property({ type: String, attribute: "aria-label" }) ariaLabel = "modal";
-    @property({ type: String, attribute: "aria-labelled-by" }) ariaLabelledBy = "modal_header";
     @property({ type: String }) ariaLabelClose = "Close Modal";
     @property({ type: String }) ariaLabelCancel = "Cancel Modal";
     @property({ type: String }) ariaLabelSubmit = "Submit Modal";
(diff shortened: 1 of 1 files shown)


web-components/src/components/modal/Modal.ts
```diff
@@ -271,8 +270,7 @@ export namespace Modal {
                   role="dialog"
                   id="${this.htmlId}"
                   class="md-modal ${classMap(this.modalContainerClassMap)}"
-                  aria-label="${this.ariaLabel}"
-                  aria-labelledby=${this.ariaLabelledBy}
```
neelabhnagaich, line 275: Why are we removing aria-labelledby here, as per the recommendation the dialog title needs to be associated using aria-labelledby.

binduN: if we add aria-labelledby it is reading everything thats in header , which includes the close button.
we need it to read only the title and exclude the close button @neelabhnagaich

binduN: most modals are already sending headerLabel, reset and station login which currently does not send headerLabe are not impacted by this change.

web-components/src/components/modal/Modal.ts
```diff
@@ -271,8 +270,7 @@ export namespace Modal {
...
+                  aria-label="${ifDefined(this.headerLabel || this.ariaLabel)}"
```
vinitkumargoel, line 273: Can you add undefined also if the both properties are not present then undefined will be activated and the aria-label will not be effective.
`aria-label="${ifDefined(this.headerLabel || this.ariaLabel || undefined)}"`

web-components/src/components/modal/Modal.ts
vinitkumargoel: I guess we should keep the labelled by as it is. Did you check are we using this property in our desktop we don't want that functionality to be broken because of this.

vinitkumargoel (approved):

web-components/src/components/modal/Modal.ts
binduN: done

vinitkumargoel (approved):

PR #1652, opened 2024-06-10
CX-10573 : fix: Focus does not move to clear input button properly

iruchipriya:
CX-10573
## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

Vidcast - https://app.vidcast.io/share/6b48c381-36bb-457a-ab02-7ca62cb994d3

### DIFF
--- a/web-components/src/components/input/Input.ts
+++ b/web-components/src/components/input/Input.ts
@@ -264,8 +264,12 @@ export namespace Input {
       }
     }
 
-    handleChange(event: Event) {
-      this.value = (event.target as HTMLInputElement).value;
+    handleChange(event: Event| MouseEvent | KeyboardEvent) {
+      if (event instanceof MouseEvent || event instanceof KeyboardEvent) {
+        this.value = '';
+      } else {
+        this.value = (event.target as HTMLInputElement).value;
+      }
       this.dispatchEvent(
         new CustomEvent("input-change", {
           bubbles: true,
--- a/web-components/src/components/input/Input.ts
+++ b/web-components/src/components/input/Input.ts
@@ -211,6 +211,13 @@ export namespace Input {
     }
 
     handleKeyDown(event: KeyboardEvent) {
+      if (event.code === 'Tab') {
+        event.preventDefault();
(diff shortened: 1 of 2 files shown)


akshatatray (approved):

web-components/src/components/input/Input.ts
```diff
@@ -292,8 +299,19 @@ export namespace Input {
         }
         event.preventDefault();
       }
+      this.input.click();
+      this.value = '';
+      this.dispatchEvent(
+        new CustomEvent("input-change", {
```
vikcool2pr, line 305: looks a duplicate code , handleChange doing same thing right

iruchipriya: updated the function to handle all conditions

web-components/src/components/input/Input.ts
```diff
@@ -443,6 +456,7 @@ export namespace Input {
                 class="md-input__icon-clear"
                 name="clear-active_12"
                 aria-label=${this.clearAriaLabel || "Clear Input"}
+                tabindex="1"
```
neelabhnagaich, line 459: why we need tabindex for md-icon as md-button is itself a focusable element

iruchipriya: we will be needing this to specify the order of focus. since after pressing tab on input, the focus gets lost and on second click only it goes to the button. so we will be needing tab index to fix that

web-components/src/components/input/Input.ts
```diff
@@ -292,8 +303,10 @@ export namespace Input {
         }
         event.preventDefault();
       }
-      this.input.focus();
+      this.input.click();
```
mithun-srinivas, line 306: why are we forcing a click here inside handleChange method ?

mithun-srinivas (commented):
Please check this

web-components/src/components/input/Input.ts
neelabhnagaich: this will result in button and icon both getting the focus, which would not be desirable

PR #1760, opened 2024-08-28
fix(tooltip): hide tooltip on scroll

lhernonCisco:
## Description

Fix for CAUI-9926. Tooltip shouldn't stay visible for element when scrolling.

## Related Issue

https://jira-eng-sjc12.cisco.com/jira/browse/CAUI-9926

## How Has This Been Tested?

Tested manually, tooltip now hides when page scrolls.

## Screenshots:
**Before** (If applicable):

TODO

**After**:

TODO

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/web-components/src/components/tooltip/Tooltip.ts
+++ b/web-components/src/components/tooltip/Tooltip.ts
@@ -60,14 +60,22 @@ export namespace Tooltip {
       this.handleKeyDown(e);
     };
 
+    private _scrollListener = (e: Event) => {
+      if (this.opened) {
+        this.notifyTooltipDestroy();
+      }
+    }
+
     connectedCallback() {
       super.connectedCallback();
       document.addEventListener("keydown", this._keyDownListener);
+      window.addEventListener("scroll", this._scrollListener);
     }
 
     disconnectedCallback(): void {
       super.disconnectedCallback();
       document.removeEventListener("keydown", this._keyDownListener);
+      window.removeEventListener("scroll", this._scrollListener);
     }
 
     protected handleFocusIn(event: Event) {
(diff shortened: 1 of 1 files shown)


web-components/src/components/tooltip/Tooltip.ts
```diff
@@ -60,14 +60,22 @@ export namespace Tooltip {
...
+    private _wheelListener = (e: Event) => {
+      if (this.opened) {
+        this.notifyTooltipDestroy();
+      }
+    };
+
     connectedCallback() {
       super.connectedCallback();
       document.addEventListener("keydown", this._keyDownListener);
+      window.addEventListener("wheel", this._wheelListener, { passive: true });
```
lhernonCisco, line 72: Screenshot 2024-09-03 at 11 04 02

I'm a little concerned about using this event listener, due to the sheer volume of events generated whenever the user scrolls. However, the logic in the listener does guard the expensive call from being made more than once, as shown with the logs above. 

The reason I'm using wheel events rather than scroll events is because the scroll events are captured by the actual element which is being scrolled. In many cases, this is often a container div rather than the window itself. As such, the tooltip may be within one or more other scrollable containers, none of which can be efficiently obtained from the tooltip itself, without doing an exhaustive search of the DOM, so far as I'm aware. Additionally, because these elements have `overflow: something` in their CSS, even setting `window.addEventListener("scroll", ()=> {...}, true)` won't get called.  

Wheel events are fired to each element, regardless of where they are in the DOM.

The other main use cases in which scroll events are generated without wheel events are largely irrelevant to this specific code change:
- keyboard navigation: in this case, on the first click of the arrow keys, the tooltip disappears. This is also the behaviour if the key is held down.
- Using the mouse on the scroll bar: in this case, the mouse cannot be hovering over an element with a tooltip
- scrolling with JavaScript: haven't tested this use case, but in general terms as soon as the scroll ends, the tooltip will disappear, so it isn't regression in any case.

https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event

vikcool2pr (dismissed):

PR #1791, opened 2024-10-03
fix: CX-16357 Adding blur event handler so the menu closes if iframe is clicked

mruane1:
## Description
Adding a blur event handler so that I can hook onto a click inside an iframe while the menu is open.
The expected behaviour of the open menu in this case would be to close.
It can be difficult to track any interaction with an embedded iframe from the parent app, but using the window's blur event will let us know when focus has moved away from the current application, and into another app or an iframe in this case.

## Related Issue

CX-16357 - When opening the filter dropdown in the Agent Performance Statistics page, the horizontal header tab are not closing

## Motivation and Context

This will allow open menus to close on outside click when the click is inside an iframe. Prior to this change, the menu would remain open as the iframe would not register a click event within the parent app.

## How Has This Been Tested?

To test this change I recreated the bug scenario in storybook by adding an iframe to the example. From here I was able to open the menu and click on the iframe while the menu was open.
I wanted to ensure that this new change impacts as little of the existing behaviour as possible so I have not changed any existing code, only added a new handler. I am also adding a conditional check to only execute the close logic if the click did in fact come specifically from an iframe.
Also tested nested menu scenario to ensure that was still working as expected.

## Screenshots:
**Before** (If applicable):

https://github.com/user-attachments/assets/fc4076df-6854-4a5b-a82c-6f943dbee19c

**After**:

https://github.com/user-attachments/assets/24a62555-33e3-4d75-9f67-a44d4ce26624

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/web-components/src/components/menu-overlay/MenuOverlay.stories.ts
+++ b/web-components/src/components/menu-overlay/MenuOverlay.stories.ts
@@ -83,6 +83,7 @@ export const MenuOverlay = () => {
           </div>
         </div>
       </md-menu-overlay>
+      <iframe style="float: left; margin-top: 200px;" title="iframe-to-test-menu-closing" src="http://127.0.0.1:8080/?path=/story/components-date-range-picker--date-range-picker"></iframe>
     </md-theme>
   `;
 };
--- a/web-components/src/components/menu-overlay/MenuOverlay.ts
+++ b/web-components/src/components/menu-overlay/MenuOverlay.ts
@@ -149,7 +149,7 @@ export namespace MenuOverlay {
     connectedCallback() {
       super.connectedCallback();
       this.handleIframeClick = this.handleIframeClick.bind(this);
-      window.addEventListener('blur', this.handleIframeClick);
+      window.addEventListener("blur", this.handleIframeClick);
       document.addEventListener("click", this.handleOutsideOverlayClick);
       document.addEventListener("keydown", this.handleOutsideOverlayKeydown);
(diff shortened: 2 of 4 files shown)


web-components/src/components/menu-overlay/MenuOverlay.stories.ts
```diff
@@ -83,6 +83,7 @@ export const MenuOverlay = () => {
           </div>
         </div>
       </md-menu-overlay>
+      <iframe style="float: left; margin-top: 200px;" title="iframe-to-test-menu-closing" src="http://127.0.0.1:8080/?path=/story/components-date-range-picker--date-range-picker"></iframe>
```
mruane1, line 86: Thought I would leave this sample iframe in the example to ensure this behaviour stays working as any other updates are made

web-components/src/components/menu-overlay/MenuOverlay.ts
```diff
@@ -148,6 +148,8 @@ export namespace MenuOverlay {
 
     connectedCallback() {
       super.connectedCallback();
+      this.handleIframeClick = this.handleIframeClick.bind(this);
```
mruane1, line 151: Needed to explicitly bind this here as the context of 'this' would otherwise be lost when iframe is clicked.

NotNestor (approved):

PR #1797, opened 2024-10-08
Origin/cx 16499 menu overlapping border fix

mruane1:
fixing the overlapping visible border on the menu by bringing the css pointer to the front and adding a border around the new pointer position

## Description
In order to create a seamless border around the menu I first moved the pointer to sit on top of the menu.
This solved the problem of the menu border showing as the pointer now covered this area.

The next step was to create a border around the pointer so that the border runs around the entire menu and pointer.
In order to do this, I created another pointer with the same color as the menu border and positioned this behind the original pointer. By offsetting the overlaying original pointer by 2 pixels, this achieved a matching border around the menu pointer.
image

Video to demonstrate:
https://github.com/user-attachments/assets/45a4a808-c4c9-4fce-9d14-fd6ee2234387

A further issue was highlighted by fixing this border issue. Something that is visible today is that the initial position of the arrow pointer is incorrect. When the menu first opens, the arrow is not where it should be. It is only after you interact with the page does the position get corrected. It was tricky to track this down but I eventually traced it back to popper.js lib. This is down to how the popper.js calculates position and how we are showing and hiding the menus. Popper.js cannot calculate exact positioning until the menu is open, so we must call .update() on the popper instance once we know the menu has been created and is open. This allows popper to calculate the position correctly in relation to the menu trigger.

See the issue below:
https://github.com/user-attachments/assets/c9d0e12f-b2f4-4dad-8ddf-74f8eb19fdef

This is now fixed. See After screen recording below.

## Related Issue

CX-16499

## Motivation and Context

The border overlapping the menu pointer did not look good. This provides a more seamless look.

## How Has This Been Tested?

I have tested this using all of the dev examples as seen in the screen recordings.
I have also pulled the local build into the web client consuming application to ensure that there were no undesirable effects from this change.

## Screenshots:
**Before** (If applicable):

image

**After**:

https://github.com/user-attachments/assets/8daaca74-b577-454f-be6e-259b467bfe20

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


web-components/src/components/menu-overlay/MenuOverlay.ts
```diff
@@ -287,6 +287,7 @@ export namespace MenuOverlay {
           onFirstUpdate: async () => {
             // We need to find all focusable elements, after Popper finish its positioning calculation
             if (this.isOpen) {
+              this.popperInstance?.update();
```
mruane1, line 290: This fixes the issue mentioned in the PR description.
Known issue: https://stackoverflow.com/questions/46476857/popper-js-wrong-initial-position-of-dropdown-tooltip

web-components/src/components/menu-overlay/scss/settings.scss
```diff
@@ -1,5 +1,5 @@
 $arrow-size: rem-calc(16);
-$arrow-offset: rem-calc(2) - $arrow-size;
+$arrow-offset: rem-calc(0.8) - $arrow-size;
```
mruane1, line 2: Arrow was sitting behind the menu prior to this change, which was essentially causing the border issue, so this value did not need to be very accurate. Adjusting this to accurately align with edge of menu.

web-components/src/components/menu-overlay/scss/menu-overlay.scss
```diff
@@ -48,28 +46,64 @@
...
+
+  .overlay-container > .overlay-arrow:after {
+    content: "";
+    position: absolute;
+    border-width: 19px;
+    border-style: solid;
+    border-color: transparent;
+  }
+
+  // styling to create triangle border to match menu for pointer
```
mruane1, line 85: Below just creates a menu pointer with the same color as the border color. This allows us to overlay the white menu pointer and offset it by 2 pixels given us an arrow with a seamless border:
image

NotNestor (approved):

PR #1822, opened 2024-11-12
fix(release): decouple title from rendered text in TextItem (CX-16559)

cormacmchale1:
This will allow users to optionally avoid the browser generated tooltip from title.

## Description

This will allow users to optionally avoid the browser generated tooltip from title. By providing a new variable to use instead of title as the display text. If title is only used. Then the legacy functionalty will reamin and it will be used as display text.

## Motivation and Context

Prevent multiple tooltips in the case where a tooltip has been added to a task-item.

## How Has This Been Tested?

Will not affect anything external to the component, or any pre-exsisting functionality on the component.
Run and verified in the web-component sandbox.
Run against the suite of tests exisiting for the component, with tests added to cover my change.

## Screenshots:
**Before** (If applicable):

Functionality is preserved in this case. This changes allows any consumer the ability to remove the broswer tooltip but retain the ability to display the visible title on screen.

**After**:

Screenshot 2024-11-13 at 09 04 54
Screenshot 2024-11-13 at 09 05 04
Screenshot 2024-11-13 at 09 05 11

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/web-components/src/components/taskitem/TaskItem.test.ts
+++ b/web-components/src/components/taskitem/TaskItem.test.ts
@@ -20,7 +20,7 @@ const fixtureFactory = async (
     <md-task-item
       mediaType="${mediaType}"
       popovertitle="${popovertitle}"
-      itemTitle="${itemTitle}"
+      item-title="${itemTitle}"
       title="${title}"
       status="${status}"
       quantity="${quantity}"
--- a/web-components/src/components/taskitem/TaskItem.test.ts
+++ b/web-components/src/components/taskitem/TaskItem.test.ts
@@ -422,12 +422,12 @@ describe("TaskItem", () => {
     expect(ariaLabel).toEqual("Custom area label");
   });
 
-  test("should have custom aria label", async () => {
+  test("should have custom aria label title", async () => {
     const element: TaskItem.ELEMENT = await fixtureFactory(
       "twitter",
       "Mihael Varificantare",
+      "Mihael Varificantare aria label",
       "",
-      "Mihael Varificantare",
       "quelle_1",
       "transfered",
       0,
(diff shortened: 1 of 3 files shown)


web-components/src/components/taskitem/TaskItem.ts
```diff
@@ -31,6 +30,37 @@ export namespace TaskItem {
...
+    @property({ type: String, reflect: true})
+    get itemTitle() {
+      return this.displayTitle;
+    }
+    set itemTitle(itemTitle: string) {
+      this.updateTitle(itemTitle);
+    }
+
+    private updateTitle(title: string) {
+      if (!title) {
```
cormacmchale1, line 56: This logic covers the cases of passing blank strings for either variable. The test component is setup in this way, seems like a legitimate case to cover

web-components/src/components/taskitem/TaskItem.test.ts
```diff
@@ -18,8 +19,9 @@ const fixtureFactory = async (
   return await fixture(html`
     <md-task-item
       mediaType="${mediaType}"
-      title="${title}"
       popovertitle="${popovertitle}"
+      itemTitle="${itemTitle}"
+      title="${title}"
```
cormacmchale1, line 24: Just to make sure the ordering of these don't break the tests with new logic

web-components/src/components/taskitem/TaskItem.ts
```diff
@@ -30,6 +29,43 @@ export namespace TaskItem {
...
+      this.itemTitleValue = itemTitle;
+      this.updateTitle();
+    }
+
+    private updateTitle() {
+      if(!this.itemTitleValue){
+      this.displayTitle = this.titleValue;
+      }
+      else{
+        this.displayTitle = this.itemTitleValue;
```
cormacmchale1, line 64: We just prioritise itemTitle for display text. But will use title if nothing is provided.

NotNestor: removing title is a breaking API change. this should be marked as deprecated and left

web-components/src/components/taskitem/TaskItem.ts
```diff
@@ -30,6 +29,43 @@ export namespace TaskItem {
...
+    }
+    set title(title: string) {
+      this.titleValue = title;
+      this.updateTitle();
+    }
+    /**
+    * The visible text for the Title.
+    */
+    @property({ type: String, reflect: true, attribute: "itemtitle" })
+    get itemTitle() {
```
NotNestor, line 51: update the attribute to item-title

NotNestor: when this is item-title you can use the attribute item-title="title" or .itemTitle="title"

web-components/src/components/taskitem/TaskItem.ts
cormacmchale1: This doesn't remove it though, sorry if I'm not following, any legacy usage is preserved in this PR.

web-components/src/components/taskitem/TaskItem.ts
```diff
@@ -30,6 +29,43 @@ export namespace TaskItem {
     @property({ type: String }) customAriaLabel = "";
     @property({ type: String }) iconSrc = "";
     @property({ type: String }) tabIndexForContainer = "0";
+    @property({ type: String }) titleValue = "";
```
cormacmchale1, line 32: These should be private. Will update

PR #1861, opened 2025-01-16
# CX-17328 - fix(release): update event flow for date picker

cormacmchale1:
Events need to propogate to the UI impl on blank values for dates. So that the UI can respond appropriately.

## Description
The if statement would ignore blank values, and the method flow would propogate the previous value, which is then ignored by DateTimePicker.ts in it's updated flow.

## Motivation and Context
Makes sure any UI that implements this can respond if required to a blank value entering the input field for the date picker.

## How Has This Been Tested?
Manually(Sandbox) and in tests.

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/web-components/src/components/date-time-picker/DateTimePicker.ts
+++ b/web-components/src/components/date-time-picker/DateTimePicker.ts
@@ -98,9 +98,7 @@ export namespace DateTimePicker {
     };
 
     handleDateTimeInputChange = (event: CustomEvent) => {
-      if (event?.detail?.value) {
-        this.value = event?.detail?.value;
-      }
+      this.value = event?.detail?.value;
     };
 
     parseValueForVisuals = (value: string) => {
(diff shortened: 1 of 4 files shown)


NotNestor (dismissed):

web-components/src/components/date-time-picker/DateTimePicker.ts
```diff
@@ -111,21 +111,27 @@ export namespace DateTimePicker {
...
-              dateTimeString: this.value,
-              dateTime: this.fullDateTime,
-              locale: this.locale,
-              twentyFourHourFormat: this.twentyFourHourFormat
-            }
-          })
-        );
       }
+      else
+      {
```
cormacmchale1, line 119: This is worth talking about. This will send a datetime back as 
Screenshot 2025-01-16 at 11 36 21

Which is correct but worth flagging

cormacmchale1: Needs to be done in this manner as typescript won't allow the use of this.value without the check.

cormacmchale1: The formatting on the code is updated since this commit

NotNestor (approved):

PR #1876, opened 2025-02-06
fix: correct menu overlay state handling on window blur event

rajeevdodda:
## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/web-components/src/components/menu-overlay/MenuOverlay.ts
+++ b/web-components/src/components/menu-overlay/MenuOverlay.ts
@@ -438,7 +438,7 @@ export namespace MenuOverlay {
 
     handleWindowBlurEvent() {
       if (this._isOpen) {
-        this.isOpen = false;
+        this.isOpen = this._isOpen;
       }
     }
 


rajeevdodda:
@NotNestor - In the current behavior, the menu overlay is hidden when the user switches between tabs. The above code change will fix this issue.

web-components/src/components/menu-overlay/MenuOverlay.ts
```diff
@@ -438,7 +438,7 @@ export namespace MenuOverlay {
 
     handleWindowBlurEvent() {
       if (this._isOpen) {
-        this.isOpen = false;
+        this.isOpen = this._isOpen;
```
mruane1, line 441: So this window blur block event was put in as a fix for: https://jira-eng-sjc12.cisco.com/jira/browse/CX-16357
This code is now saying if _isOpen is true, set isOpen to true. This is essentially reverting the fix for above, which if that is the course of action we are taking, we may as well remove the entire blur event.

Is this in relation to the outreach menu not staying open when moving to a new browser tab or window?
I believe a solution we were thinking of is to have a keepOpen attribute on the menu, so that for some instances of a dropdown menu, we could keep open if users navigate away from the app, but for majority of cases they will close.

web-components/src/components/menu-overlay/MenuOverlay.ts
```diff
@@ -438,7 +438,11 @@ export namespace MenuOverlay {
 
     handleWindowBlurEvent() {
```
mruane1, line 439: Some redundant checks here can we clean up the function slightly. We don't need to reset isOpen to this._isOpen as it's already true if it enters the first if.
```
handleWindowBlurEvent() {
  if (this._isOpen && !this.keepOpenOnWindowBlur) {
    this.isOpen = false;
  }
}
```

mruane1:
You might already be working them but you might add unit tests to cover new scenario.

web-components/src/components/menu-overlay/MenuOverlay.ts
rajeevdodda: changed this logic as per the above suggestion

NotNestor (approved):

mruane1 (approved):

PR #1908, opened 2025-03-31
Tab accessibility fixes

NotNestor:
## Description
Accessibility updates pt 1.

This will fix the issue with md-tab having an aria hidden button to render content and manage some of the events.

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


web-components/src/[sandbox]/examples/tabs.ts
```diff
@@ -35,29 +35,29 @@ const tabsOverlayHtmlList = ["All templates", "Only Fb Template", ...Array(20)].
   `
 );
 
+const tabMenuStyle = css`
+  .menu-trigger-button {
```
Comment (NotNestor) [inline web-components/src/[sandbox]/examples/tabs.ts:39]: move common css in sandbox into reusable var

web-components/src/components/tabs/Tabs.ts
```diff
@@ -1210,34 +1259,59 @@ export namespace Tabs {
...
+              </md-tab>
+            `
+          )}
+        </div>
+        ${this.moreMenuTemplate}
+      `;
+    }
+
     render() {
       return html`
```
NotNestor, line 1316: split out the render() function template into separate logical propeties to make it easier to follow

web-components/src/components/tabs/Tabs.ts
```diff
@@ -1210,34 +1259,59 @@ export namespace Tabs {
...
     }
 
+    private get renderTabSlot() {
+      return html`<slot
+        name="tab"
+        class="${classMap({
+          "visible-tabs-slot": this.direction === "horizontal"
+        })}"
+      ></slot> `;
+    }
```
NotNestor, line 1275: in the case of the horizontal tabs this is added and then hidden. It is used to create new md-tabs from their properties. this should be revisited as it can cause issues

web-components/src/components/tabs/Tabs.ts
```diff
@@ -1210,34 +1259,59 @@ export namespace Tabs {
...
+            (tab) => html`
+              <md-tab
+                .closable="${tab.closable}"
+                .disabled="${tab.disabled}"
+                .selected="${tab.selected}"
+                name="${tab.name}"
+                id="${this.getCopyTabId(tab)}"
+                aria-label=${tab.ariaLabel}
+                aria-controls="${this.getAriaControlId(tab)}"
+                .isCrossVisible=${true}
```
NotNestor, line 1299: There is still an issue with this approach in that the copy resets the aria controls but they are now in separate shadow doms

web-components/src/components/tabs/scss/tab.scss
```diff
@@ -6,14 +6,14 @@
   box-shadow: none !important;
   outline: none;
 
-  button {
+  .tab-content {
     outline: $tab-main-focus-border;
```
NotNestor, line 10: the inner element is now a div instead of a button so css needed updating.

In general it is better to style by class rather than type so this kind of change isn't necessary

AlanSpillane (approved):

PR #2053, opened 2025-10-03
feat(md-input): another tweak to the character-count label behaviour

leonardjoe:
## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:


web-components/src/components/input/Input.ts
```diff
@@ -614,10 +614,13 @@ export namespace Input {
     }
 
     private characterCountLabelTemplate() {
-      return this.maxLength && this.displayCharacterCount && !this.disabled && !this.readOnly
```
leonardjoe, line 617: Tweaking this again, the ask is to have this label work more as a "suggested" limit rather than a hard limit. So the user can still enter more characters, but we show an error message beneath the box if they do.

PR #2138, opened 2026-05-05
#CX-24777: Fix-Momentum md-table-advanced error internal state property not resetting in updated event

gvallath:
…ges, so a previous invalid dataset does not keep blocking table rendering after valid data is provided. Made changes to the update event  to reset the error state internal property

## Description
Fix:-Momentum md-table-advanced error internal state property not resetting in updated event
Clear stale md-table-advanced data errors when the data property changes, so a previous invalid dataset does not keep blocking table rendering after valid data is provided.

Changes

Reset this.error inside update() when data changes.
Added a regression test that:
renders invalid table data and confirms the error-only state is shown
updates the component with valid data
verifies the stale error is cleared
verifies the table renders again

## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/web-components/src/components/table-advanced/TableAdvanced.ts
+++ b/web-components/src/components/table-advanced/TableAdvanced.ts
@@ -77,10 +77,16 @@ export namespace TableAdvanced {
       this.populateTable();
     }
 
+    protected willUpdate(changedProperties: PropertyValues): void {
+      super.willUpdate?.(changedProperties);
+      if (changedProperties.has("data") && changedProperties.get("data") !== undefined) {
+        this.error = "";
+      }
+    }
+
     protected update(changedProperties: PropertyValues) {
       super.update(changedProperties);
       if (changedProperties.has("data")) {
-        this.error = "";
         this.updateTableData();
         this.updateDataInTable();
       }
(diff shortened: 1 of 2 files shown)


web-components/src/components/table-advanced/TableAdvanced.ts
```diff
@@ -80,6 +80,7 @@ export namespace TableAdvanced {
     protected update(changedProperties: PropertyValues) {
       super.update(changedProperties);
       if (changedProperties.has("data")) {
+        this.error = "";
         this.updateTableData();
```
NotNestor, line 90: this should be moved to will update

willUpdate(changedProperties) — Runs before rendering. Setting reactive properties here is included in the current render cycle and does not trigger an additional update.
update(changedProperties) (after super.update()) — Rendering has already happened. Setting a reactive property here triggers a second render cycle, which is wasteful.
So if this.type is a reactive property (@property or @state) and it's being computed/reset based on changed properties, it should be in willUpdate()

gvallath: Added the willUpdate() life cycle event.

NotNestor (approved):

NotNestor (approved):

PR #2167, opened 2026-08-31
fix: prevent focus from escaping full-screen floating modal

suhanii21:
## Screenshots:
**Before** (If applicable):

**After**:

## Types of changes

- [ ] Documentation update
- [ ] Build related changes
- [ ] CI related changes

## Checklist:

### DIFF
--- a/web-components/src/components/floating-modal/FloatingModal.test.ts
+++ b/web-components/src/components/floating-modal/FloatingModal.test.ts
@@ -630,6 +630,35 @@ describe("Floating Modal Component", () => {
     expect(element.shadowRoot?.activeElement).toBe(firstHeaderButton);
   });
 
+  test("should reach content added to the panel after opening, with no external refresh signal", async () => {
+    element.show = true;
+    element.full = true;
+    jest.advanceTimersByTime(600);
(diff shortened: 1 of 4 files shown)


suhanii21:
BUILD

NotNestor:
Removing this.focusableElements = [] avoids temporarily disabling the focus trap, but it introduces a race in the shared FocusTrapMixin. During the 10 ms debounce, the cached list may contain controls that have been removed or hidden after an on-widget-update event. If Tab is pressed in that window, the event is prevented and the trap may try to focus a stale element, causing focus to become stuck or skip controls. This can affect all mixin consumers, including modal, popover, menu overlay, coachmark, and advanced table.
Could we preserve trapping without trusting stale targets? For example, recompute the list synchronously when handling Tab, or filter cached entries for connectivity, visibility, and focusability before selecting the next target. A regression test should remove/hide a focusable element, dispatch on-widget-update, and press Tab before the debounce completes.

suhanii21:
BUILD

suhanii21:
BUILD

NotNestor (approved):

